PR #6, opened 2017-10-08
make setup role thing allow multi-worded things, renamed modmail to Mod Mail

fourjr:
cool stuff :)


verixx:
can you make `Mod Mail` into `mod mail`. Just for consistancy, im pretty sure everything gets converted to lowercase, but just want to be safe

PR #42, opened 2018-06-24
add on_command_error handler

bananaboy21:
so the bot won't stay silent if a non-admin tries to use setup or disable


fourjr:
yeah you can do that

SharpBit:
Lol both of u r lazy

PR #98, opened 2019-01-05
Heroku documentation .env vars

iamluca:
Forgot in the heroku instance to put the prefix.

### DIFF
--- a/app.json
+++ b/app.json
@@ -8,8 +8,12 @@
         "required": true
     },
     "GUILD_ID": {
-        "description": "The id for the server you are hosting this bot for.",
+        "description": "The id for the server you are serving this bot for.",
         "required": true
+    },
+    "MODMAIL_GUILD_ID": {
+        "description": "Operations center for the server you are serving for.",
+        "required": false
     },
     "OWNERS": {
         "description": "Comma seperated user IDs of people that are allowed to use owner only commands. (eval and update)",
--- a/app.json
+++ b/app.json
@@ -12,7 +12,7 @@
         "required": true
     },
     "MODMAIL_GUILD_ID": {
-        "description": "Operations center for the server you are serving for.",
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -8,13 +8,21 @@
...
+        "description": "Operations center for the server you are serving for.",
+        "required": false
+    },
     "OWNERS": {
         "description": "Comma seperated user IDs of people that are allowed to use owner only commands. (eval and update)",
         "required": true
     },
+    "PREFIX": {
+        "description": "Beginning of command prefix.",
+        "required": true
```
kyb3r: Should not be required, also document that the default prefix is `?`

app.json
```diff
@@ -8,13 +8,21 @@
         "required": true
     },
     "GUILD_ID": {
-        "description": "The id for the server you are hosting this bot for.",
+        "description": "The id for the server you are serving this bot for.",
         "required": true
     },
+    "MODMAIL_GUILD_ID": {
+        "description": "Operations center for the server you are serving for.",
```
kyb3r: Describe that this config var is only necessary if you want the bot to have a separate operations center.

app.json
iamluca: Ok. Fixing

app.json
iamluca: Fixing

iamluca:
@kyb3r Check now!

app.json
kyb3r: Try: `The server which is the operations center for the bot. This is only necessary if you want a separate server in which modmail operations occur.`

app.json
kyb3r: You forgot to say that the default prefix is `?`

kyb3r:
Config that is not required at start time should not be in here. You can easily set up a prefix, status, etc with the commands within the bot.

PR #119, opened 2019-01-10
Allow for "Custom response message for thread creation"

CleverSource:
The configuration variable name is kind of weird, no clue if that should be changed. I don't know if I used self.bot.config.get() correct either but based on the code it takes in 2 parameters for the configuration variable to get, and if that configuration variable isn't set it picks a default or the 2nd parameter.

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -8,7 +8,7 @@ class ConfigManager:
 
     allowed_to_change_in_command = {
         'status', 'log_channel_id', 'mention', 'disable_autoupdates', 'prefix',
-        'main_category_id', 'sent_emoji', 'blocked_emoji'
+        'main_category_id', 'sent_emoji', 'blocked_emoji', 'response_thread_created'
         }
     
     internal_keys = {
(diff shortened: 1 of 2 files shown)


core/config.py
```diff
@@ -8,7 +8,7 @@ class ConfigManager:
 
     allowed_to_change_in_command = {
         'status', 'log_channel_id', 'mention', 'disable_autoupdates', 'prefix',
-        'main_category_id', 'sent_emoji', 'blocked_emoji'
+        'main_category_id', 'sent_emoji', 'blocked_emoji', 'response_thread_created'
```
kyb3r: maybe change the name to `thread_creation_response`

CleverSource: Much better config name. 👍

kyb3r:
I just realised, we should increment the `MINOR` version when adding backwards compatible added functionality. So can you change the version to `2.1.0`

kyb3r:
You may need to store `status_type` in the config as well so when the bot starts it will set the correct status type. (`on_connect` in `bot.py`)

iamluca:
Can push now.

kyb3r:
I don't think in its current state the status command will work.

CleverSource:
I'm not fond of the new discord rewrite, I completely broke the status command and haven't really debugged how to fix it.

PR #129, opened 2019-01-14
Add ability to selfhost logs 

kyb3r:
This pr implements a separate mode in the bot which makes it able to store logs in a user-owned database. A mongodb cluster uri needs to be inputted for this mode to be enabled. 

To enable this mode you need to add the following config variables on Heroku.
`MONGO_URI` - The mongodb cluster connection URI 
`LOG_URL` - The URL of your Heroku app. e.g. `https://yourappname.herokuapp.com`

You will also need to turn on the web worker.

### DIFF
--- a/app.py
+++ b/app.py
@@ -14,11 +14,11 @@ async def init(app, loop):
 
 @app.get('/')
 async def index(request):
-    return response.text('Welcome! This simple webserver is used to display your modmail logs.')
+    return response.text('Welcome! This simple website is used to display your modmail logs.')
 
 @app.get('/logs/<key>')
 async def getlogsfile(request, key):
-    '''Returned the plain text rendered log entry'''
+    """Returned the plain text rendered log entry"""
 
     log = await app.db.logs.find_one({'key': key})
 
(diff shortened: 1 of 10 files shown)


app.py
```diff
@@ -0,0 +1,31 @@
...
+async def init(app, loop):
+    app.db = AsyncIOMotorClient(os.getenv('MONGO_URI')).modmail_bot
+
+@app.get('/')
+async def index(request):
+    return response.text('Welcome! This simple webserver is used to display your modmail logs.')
+
+@app.get('/logs/<key>')
+async def getlogsfile(request, key):
+    print(key)
```
fourjr: Print?

core/models.py
```diff
@@ -0,0 +1,74 @@
+from datetime import datetime
+import dateutil.parser
+
+class LogEntry:
+    def __init__(self, data):
+        self.key = data['key']
+        self.open = data['open']
+        self.created_at = dateutil.parser.parse(data['created_at'])
```
fourjr: Don't use dateutil.parser

PR #131, opened 2019-01-15
Added `activity` command for Issue #116

### LINKED ISSUE #116: Support multiple bot presence activities
İ think add bot a this property.
We can change bot status watching and listening. 
<h2> Example </h2> 

*Watching* DM me for help!
*Listening* DM me for help!

<h2>Example Command</h2>

!status-watching "Status"
!status-listening "Status"

Taaku18:
Fixes #116 

Added an `activity` command for setting the activity of ModMail, usage:

  - `[prefix]activity activity_type [message...]`
  - `[prefix]presence  activity_type [message...]`

Activity types can be: `playing`, `streaming`, `listening`, `watching`, or `clear`.
If activity type is `clear`, the current activity is removed, and no `message` is required.

## Notes:

  - Did **not** use `status` as the command name since discord rewrite reserves "status" for the state of the bot. Ie `online`, `offline`, `idle`, `dnd`, and `invisible`.
  - Slightly updated `.gitignore`

### DIFF
--- a/cogs/utility.py
+++ b/cogs/utility.py
@@ -267,28 +267,47 @@ class Utility:
             else:
                 em.description = 'Already up to date with master repository.'
 
-            
         await ctx.send(embed=em)
 
     @commands.command(aliases=['presence'])
     @commands.has_permissions(administrator=True)
-    async def status(self, ctx, activity_type: str, *, message: str = ''):
+    async def activity(self, ctx, activity_type: str, *, message: str = ''):
         """
-        Set a custom playing status for the bot.
+        Set a custom activity for the bot.
 
-        [prefix]status activity message...
+        Possible activity types: `playing`, `streaming`, `listening`, `watching`, `clear`
 
-        Set the message to `clear` if you want to remove the status.
+        When activity type is set to `clear`, the current activity is removed.
         """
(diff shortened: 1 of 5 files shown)


cogs/utility.py
```diff
@@ -265,30 +267,60 @@ def format_not_found(self, ctx, command):
...
         await ctx.send(embed=em)
 
-    @commands.command(name='status', aliases=['customstatus', 'presence'])
+    @commands.command(aliases=['presence'])
     @commands.has_permissions(administrator=True)
-    async def _status(self, ctx, *, message):
-        """Set a custom playing status for the bot.
-
-        Set the message to `clear` if you want to remove the playing status.
+    async def activity(self, ctx, activity_type: str, *, message: str = ''):
```
kyb3r: Why not remove the message default value and let the command handler output the usage of the command.

Taaku18: The default's for the `clear` activity type to remove the current activity, I don't think `clear` will work if the default's removed.

kyb3r: Yes, the command error handler checks if that error was raised and sends the help message for that command.

Taaku18: Updated.

PR #135, opened 2019-01-17
Scheduled close event for thread save to config vars

### LINKED ISSUE #134: Use config vars to store scheduled closure time
> Currently scheduled closing uses asyncio sleep but since heroku apps restart every 24 hours it wont work for closing for more than 24 hours so i'm thinking of storing the scheduled close date in the config and on bot start try sleeping until the close date.
> \- Kyber

Currently, my approach:

> So I'm thinking like having a config var `closures` and set it to a  mapping (thread id-> thread closure time [in iso-8601 format]) . We can update the config var when there's new there's a new scheduled thread closure and retrieve them and add them back to cache when the bot starts.

I’ll be working on this over the next few days.

Taaku18:
According to [Automatic dyno restarts](https://devcenter.heroku.com/articles/dynos#automatic-dyno-restartsl), all dynos will restart at least once every 27 hours. 

> Dynos are also restarted (cycled) at least once per day to help maintain the health of applications running on Heroku. Any changes to the local filesystem will be deleted. The cycling happens once every 24 hours (plus up to 216 random minutes, to prevent every dyno for an application from restarting at the same time). 

During this period, local caches are deleted, which results in the inability to set the scheduled closure time to longer than 24 hours.
 
---

This PR resolves this issue.
#Fixes #134 

 - Use `call_later()` instead of `sleep()` for schelduling.
 - Created a new internal config var: `closures`.
 - Bot:
    - loads `closures` upon start
    - removes `closures['id']` upon deletion 

Format: 

    'closures': {
        'recipient id': {
            'time': 'ISO 8601 time string',
            'closer_id': int,
            'silent': bool,
            'delete_channel': bool,
            'message': 'message string'
        },
        ...
    }

Unresolved Known Issues
-------------------------
  - During the bot's downtime (when bot restarts), messages during this period are ignored. Therefore the close schedule will not terminate.

---

# Aside

#Addresses #133 

 - Improved formatting in core/config.py

---

**Please carefully review this PR and minor version bump upon approval.**

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -204,7 +204,7 @@ class ModmailBot(commands.Bot):
             await self.threads.populate_cache()
         await self.config.update()
 
-        closures = self.config.get('closures', {})
+        closures = self.config.get('closures', {}).copy()
         print(closures)  # for debugging
         for recipient_id, items in closures.items():
             after = (datetime.datetime.fromisoformat(items['time']) -
@@ -212,10 +212,18 @@ class ModmailBot(commands.Bot):
             if after < 0:
                 after = 0
             recipient = self.get_user(int(recipient_id))
-            print(recipient)
+
             thread = await self.threads.find(
                 recipient=recipient)
-            print(thread)
+
+            if not thread:
+                # If the recipient is gone or channel is deleted
+                closures = self.config.get('closures', {})
+                closures.pop(str(recipient_id))
+                self.config['closures'] = closures
+                await self.config.update()
+                continue
+
             # TODO: Retrieve messages/replies when bot is down, from history?
             await thread.close(closer=self.get_user(items['closer_id']),
                                after=after,
--- a/bot.py
+++ b/bot.py
@@ -204,6 +204,7 @@ class ModmailBot(commands.Bot):
             await self.threads.populate_cache()
 
         closures = self.config.get('closures', {})
+        print(closures)  # for debugging
         for recipient_id, items in closures.items():
             after = (datetime.datetime.fromisoformat(items['time']) -
                      datetime.datetime.utcnow()).total_seconds()
--- a/bot.py
+++ b/bot.py
@@ -203,7 +203,21 @@ class ModmailBot(commands.Bot):
         else:
             await self.threads.populate_cache()
 
+        closures = self.config.get('closures', {})
+        for recipient_id, items in closures.items():
+            after = (datetime.datetime.fromisoformat(items['time']) -
+                     datetime.datetime.utcnow()).total_seconds()
+            if after < 0:
+                after = 0
+            thread = await self.threads.find(
+                recipient=self.get_user(recipient_id))
 
+            # TODO: Retrieve messages/replies when bot is down, from history?
+            await thread.close(closer=self.get_user(items['closer_id']),
+                               after=after,
+                               silent=items['silent'],
+                               delete_channel=items['delete_channel'],
+                               message=items['message'])
 
     async def process_modmail(self, message):
         """Processes messages sent to the bot."""
--- a/bot.py
+++ b/bot.py
@@ -202,6 +202,7 @@ class ModmailBot(commands.Bot):
             print(Fore.RED + Style.BRIGHT + 'WARNING - The GUILD_ID provided does not exist!' + Style.RESET_ALL)
         else:
             await self.threads.populate_cache()
+        await self.config.update()
 
         closures = self.config.get('closures', {})
         print(closures)  # for debugging
--- a/bot.py
+++ b/bot.py
@@ -211,7 +211,7 @@ class ModmailBot(commands.Bot):
                      datetime.datetime.utcnow()).total_seconds()
             if after < 0:
                 after = 0
-            recipient = self.get_user(recipient_id)
+            recipient = self.get_user(int(recipient_id))
             print(recipient)
             thread = await self.threads.find(
                 recipient=recipient)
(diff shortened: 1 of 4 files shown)


core/thread.py
```diff
@@ -42,28 +42,61 @@ def ready(self, flag):
...
+            return
 
+        return await self._close(closer, silent, delete_channel, message)
+
+    async def _close(self, closer, silent=False, delete_channel=True,
+                     message=None, scheduled=False):
         del self.manager.cache[self.id]
+
+        closures = self.bot.config.get('closures', {})
+        closures.pop(str(self.id))
```
kyb3r: dict.pop will raise a key error if the id is not in there, if the thread is getting closed immediately, it wont be in the closures dict right?

core/thread.py
```diff
@@ -42,28 +42,61 @@ def ready(self, flag):
...
+        self.bot.config['closures'] = closures
+        await self.bot.config.update()
+
         if str(self.id) in self.bot.config.subscriptions:
```
kyb3r: Try doing something like this instead

core/config.py
```diff
@@ -38,7 +42,7 @@ def populate_cache(self):
             'aliases': {},
             'blocked': {},
             'notification_squad': {},
-            'subscriptions': {}
+            'subscriptions': {},
```
kyb3r: Add closures here as well in the initial schema so that you won't need to do `self.config.get('closures', {})

core/thread.py
Taaku18: Yes, I just didn't want the semi-pointless `scheduled` parameter. Though for `pop`, I forgot to add the default `closures.pop(str(self.id), None)`. I agree nevertheless, that using `scheduled` will be more efficient, will change.

Taaku18: I believe it's resolved now.

bot.py
```diff
@@ -202,8 +202,32 @@ def prefix(self):
...
+            thread = await self.threads.find(
+                recipient=recipient)
+
+            if not thread:
+                # If the recipient is gone or channel is deleted
+                self.config.closures.pop(str(recipient_id))
+                await self.config.update()
+                continue
+
+            # TODO: Retrieve messages/replies when bot is down, from history?
```
kyb3r: Downtime is only a few seconds anyways I doubt this will be a problem, low priority.

bot.py
```diff
@@ -202,8 +202,32 @@ def prefix(self):
...
+            await thread.close(closer=self.get_user(items['closer_id']),
```
kyb3r: loop.create_task() maybe for performance?

core/thread.py
```diff
@@ -73,14 +103,16 @@ def ready(self, flag):
...
 
         if self.bot.selfhosted:
             log_url = f'{self.bot.config.log_url}/logs/{log_data["key"]}'
         else:
-            log_url = f"https://logs.modmail.tk/{log_data['user_id']}/{log_data['key']}" 
+            log_url = f"https://logs.modmail.tk/" \
+                      f"{log_data['user_id']}/{log_data['key']}"
 
-        user = self.recipient.mention if self.recipient else f'`{self.id}`'
+        user = self.recipient.mention if self.recipient else str(self.id)
```
kyb3r: Why remove the backticks?

Taaku18: mb thought it was a double quote 😅

core/thread.py
```diff
@@ -42,28 +42,58 @@ def ready(self, flag):
...
 
+        return await self._close(closer, silent, delete_channel, message)
+
+    async def _close(self, closer, silent=False, delete_channel=True,
+                     message=None, scheduled=False):
         del self.manager.cache[self.id]
+
+        if scheduled:
+            self.bot.config.closures.pop(str(self.id), None)
+            await self.bot.config.update()
```
kyb3r: This coro is already called further down in code, so there's no need to call it here.

cogs/modmail.py
```diff
@@ -174,11 +174,11 @@ def obj(arg):
         silent = str(message).lower() in {'silent', 'silently'}
         cancel = str(message).lower() == 'cancel'
 
-        if cancel and thread.close_task is not None and not thread.close_task.cancelled():
-            thread.close_task.cancel()
-            await ctx.send(embed=discord.Embed(color=discord.Color.red(), description='Scheduled close has been cancelled.'))
-            return
-        elif cancel:
+        if cancel:
+            if thread.close_task is not None:
```
kyb3r: You don't need to check this since the function already checks it.

bot.py
Taaku18: Idrk how to implement this, can you suggest a change?

core/thread.py
Taaku18: The referred code's outdated, don't believe there's anything to do with the current code.

bot.py
kyb3r: `self.loop.create_task(thread.close(...))` instead of `await thread.close(...)`

cogs/modmail.py
Taaku18: I only reorganized this part. I think it's meant to check if the command is `close cancel`.

 - If there is a pending close task, close it. 
 - If not, output an error message.

bot.py
Taaku18: Implemented.

kyb3r:
Update changelog, increment minor patch version, and then I will merge.

Taaku18:
> Update changelog, increment minor patch version, and then I will merge.

Ready to merge.

PR #153, opened 2019-01-24
Anon cmd and internal message logging

### LINKED ISSUE #107: The ability to be anonymous in support tickets
Perhaps adding a command `anonymous`?
In doing so it replaces your name and profile picture with the name and profile picture of the server (example taken from bot)
image

### LINKED ISSUE #147: auto config option for note commands
This is so we dont need to execute note command everytime moderators want to have internal discussion taken place in tickets. Sometimes tickets that are raised by recpients can contain topics that would involve stafff to discuss wether the recipients points are valid are not. 

Its pretty common for moderators to have internal discussion in mod tickets rather than a staff channel to keep it relevent. 

Enable auto note would mean we dont need to use note command everytime.

kyb3r:
Adds an anonymous command and now the bot logs all messages sent in thread channels. This closes #107 and closes #147

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -362,8 +362,15 @@ class ModmailBot(commands.Bot):
             cmd = message.content[len(prefix):].strip()
             if cmd in self.snippets:
                 message.content = f'{prefix}reply {self.snippets[cmd]}'
+    
+        ctx = await self.get_context(message)
+        if ctx.command:
+            return await self.invoke(ctx)
 
-        await self.process_commands(message)
+        thread = await self.threads.find(channel=ctx.channel)
+
+        if thread is not None:
+            await self.modmail_api.append_log(message, type='internal')
     
     async def on_guild_channel_delete(self, channel):
         if channel.guild != self.modmail_guild:
--- a/bot.py
+++ b/bot.py
@@ -22,7 +22,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
 
-__version__ = '2.9.4'
+__version__ = '2.10.0'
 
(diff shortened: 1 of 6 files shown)


bot.py
```diff
@@ -362,8 +362,15 @@ def recipient_color(self):
...
+    
+        ctx = await self.get_context(message)
+        if ctx.command:
+            return await self.invoke(ctx)
+
+        thread = await self.threads.find(channel=ctx.channel)
 
-        await self.process_commands(message)
+        if thread is not None:
+            await self.modmail_api.append_log(message, type='internal')
```
Taaku18: Even though there isn't a command being invoked here, the message should still propagate through discord. Or else it might mask issues in the future.

```suggestion
await self.process_commands(message)
```

cogs/modmail.py
```diff
@@ -376,15 +375,25 @@ def obj(arg):
         ctx.message.content = msg
         thread = await self.bot.threads.find(channel=ctx.channel)
         if thread:
+            await ctx.trigger_typing()
```
fourjr: Why not `async with ctx.typing():`?

kyb3r: There's no need for extra indentation, an asynchronous context manager would be suitable for long running operations.

fourjr: I just feel that it'll be typing after message sent

kyb3r: That's not how discord typing works, typing stops after a message is sent or 10 seconds.

bot.py
Taaku18: I mean add it at the end of `on_message`, that "suggested change" section might be a bit misleading.

kyb3r: But we are not logging command messages

PR #163, opened 2019-01-29
Update modmail.py

iamluca:
Fixing some localization issues.


Taaku18:
Not very sure why this change is necessary?

PR #168, opened 2019-02-07
Fixed typo #v2.13.0

HeresHavi:
# v2.13.0

### Typo
- Have a niche feature request that you think only your server would benefit from? Plugins are your go-to!

### Fixed
- Have a nice feature request that you think only your server would benefit from? Plugins are your go-to!


kyb3r:
Niche is an actual word, in this case, you would interpret it as "specialized".

PR #172, opened 2019-02-10
Enhance thread creation performance

kyb3r:
This pull request makes `thread.create` synchronous so that the first message sent can be queued to be sent as soon as a thread is created. This fixes a problem where if multiple messages are sent in quick succession, the first message sent (which triggers the thread creation) is not sent in order.


core/thread.py
```diff
@@ -277,17 +358,21 @@ def _close_after(self, closer, silent, delete_channel, message):
...
 
         if not from_mod and not note:
-            await self.bot.api.append_log(message, self.channel.id)
+            self.bot.loop.create_task(
+                self.bot.api.append_log(message, self.channel.id)
+            )
 
         if not self.ready:
-            await self.wait_until_ready()
+            self.wait_until_ready()
```
Taaku18: `wait_until_ready` is a coroutine, but never awaited? https://github.com/kyb3r/modmail/pull/172/files#diff-02108bdf20ae302859bd16f5e0c1dd1bR41

core/decorators.py
```diff
@@ -4,6 +4,12 @@
 from discord.ext import commands
 
 
+async def ignore(coro):
```
Taaku18: Not a big fan of except all clauses, but make it marginally better with `except Exception:` for PEP 8. Since this isn't really a decorator, maybe move it to utils.py?

PR #173, opened 2019-02-11
Logging

Taaku18:
### Fixed
- Will no longer show  `Unclosed client session` and `Task was destroyed but it is pending!` when the bot terminates.

 ### Changed
- `print` is replaced by logging.

---
See updated changelog for more info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,84 +25,35 @@ wheels/
 .installed.cfg
 *.egg
 
-# PyInstaller
-#  Usually these files are written by a python script from a template
-#  before PyInstaller builds the exe, so as to inject date/other infos into it.
-*.manifest
-*.spec
+
+# ---- Not in Use ----
 
 # Installer logs
-pip-log.txt
-pip-delete-this-directory.txt
-
-# Unit test / coverage reports
-htmlcov/
-.tox/
-.coverage
-.coverage.*
-.cache
-nosetests.xml
-coverage.xml
-*.cover
-.hypothesis/
+# pip-log.txt
+# pip-delete-this-directory.txt
 
 # Translations
-*.mo
-*.pot
-
-# Django stuff:
-*.log
-local_settings.py
-
-# Flask stuff:
-instance/
-.webassets-cache
-
-# Scrapy stuff:
-.scrapy
+# *.mo
+# *.pot
 
 # Sphinx documentation
-docs/_build/
-
-# PyBuilder
-target/
-
-# Jupyter Notebook
-.ipynb_checkpoints
-
-# pyenv
-.python-version
-
-# celery beat schedule file
-celerybeat-schedule
-
-# SageMath parsed files
-*.sage.py
+# docs/_build/
 
 # dotenv
-.env
+# .env
+
+# ---- End Not in Use ----
+
 
 # virtualenv
 .venv
 venv/
 ENV/
 
-# Spyder project settings
-.spyderproject
-.spyproject
-
-# Rope project settings
-.ropeproject
-
 # mkdocs documentation
-/site
+# /site
 
-# mypy
-.mypy_cache/
-*.json
-!app.json
-
-# Pycharm
+# PyCharm
 .idea/
 
 # MacOS
@@ -114,3 +65,4 @@ ENV/
 # Modmail
 config.json
 plugins/
+temp/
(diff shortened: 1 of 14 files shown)


fourjr:
Logging level should be allowed to change in the config

Taaku18:
I don’t really see the point of saving the logging level in the configurations variables. At the very most, it can be changed from the environment variable in Heroku.

fourjr:
In case, during debug, we want to use `.debug`

Taaku18:
Yeah, I do agree it should be changeable. But only through Heroku settings and environment variables, since it really isn’t necessary for an average user to set this with `?config`. It’ll be like `TOKEN` which it’s only settable through ev’s.

Taaku18:
> Allow for log level to be set in ConfigManager

Done, it's now part of the protected keys, settable through ev's.

bot.py
```diff
@@ -63,6 +74,11 @@ def __init__(self):
         self._config = None
         self._db = None
 
+        log_level = logging._nameToLevel.get(self.config.log_level.upper())
+        if log_level is not None:
+            logger.setLevel(log_level)
+            ch.setLevel(log_level)
+
```
fourjr: Maybe print a message to say it has been set or it hasn't been set

Taaku18: Added:

    bot.py - INFO: -------------------------
    bot.py - INFO: Logging level: FATAL
    bot.py - INFO: -------------------------

or 

    bot.py - INFO: -------------------------
    bot.py - INFO: Invalid logging level set.   [this line is red]
    bot.py - INFO: Using default logging level: INFO
    bot.py - INFO: -------------------------

bot.py
```diff
@@ -45,9 +45,20 @@
...
 from core.config import ConfigManager
+from core.utils import info, error
 from core.models import Bot
 from core.thread import ThreadManager
 
+
+logger = logging.getLogger('Modmail')
+logger.setLevel(logging.INFO)
+
+ch = logging.StreamHandler()
```
fourjr: Specify `sys.stdout` as an arg.

```
If stream is specified, the instance will use it for logging output; otherwise, sys.stderr will be used.
```

fourjr:
You might want to properly print the traceback of `on_command_error` as well (while ignoring `commands.CommandNotFound`)

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# [Unreleased]
+
+### Added
+- You will no longer need to view your bot debug logs from Heroku. `?debug` will show you the you the recent logs within 24h through a series of embeds.
+  - If you don't mind your data (may or may not be limited to: user ID, guild ID, bot name) be on the internet, `?debug hastebin` will upload a formatted logs file to https://hastebin.com.
```
fourjr: dont put prefix in the changelog

temp/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Please disregard this directory, it is used to store Modmail temp files.
+# DO NOT REMOVE THIS DIRECTORY.
+*
+!.gitignore
```
fourjr: why don't you just git ignore the whole dir, and create it on bot start?

Taaku18: True, also a valid option.

Taaku18: Regardless, it's created on start now.

Taaku18:
# Current views of the debug command:
debug
debug hastebin
debug clear
## In hastebin:
in hastebin.com
---
There should be no distracting changes in Heroku logs and locally deployed bot consoles.

.gitignore
```diff
@@ -25,84 +25,35 @@ wheels/
 .installed.cfg
 *.egg
 
-# PyInstaller
-#  Usually these files are written by a python script from a template
-#  before PyInstaller builds the exe, so as to inject date/other infos into it.
-*.manifest
-*.spec
+
+# ---- Not in Use ----
```
fourjr: Don't need to delete them, this is a standard python git ignore

Taaku18: Lol but these are the ignores we definitely won't ever use for Modmail.

Taaku18: I Added them back according to https://github.com/github/gitignore/blob/master/Python.gitignore

PR #195, opened 2019-03-13
Custom Haste Support

ionadev:
# Support For Custom Hastebin

* As hastebin is [open-source](https://github.com/seejohnrun/haste-server) many people, ( like me ) have their own hosted hastebin's.
* This PR Gives The Ability To Users To Support Their Own hastebin. To Do this, The User has to add an enviornment variable `HASTE_URL` with its value to the haste's url. The Examples are attached Below.

### Images
---
* Enviornment Variable Example - 
image

* Working Example - 
image

* When No Enviornment Variable Is Set - 
image


ionadev:
## P.S.

* Ignore So many Commits, There are Only 2 files changes - 
image

Issue #191, opened 2019-03-05
excessive amount of roles causes issues for channel creation.


maidos:
When a user have too many roles the bot are unable to create a channel due to discord internal limitation of 2000 character per message.
Because mod mail first message during thread creation includes roles of the recipient the bot are unable to create the actual channel but the thread gets created instead.

This causes problems where theres no channel and you cant close the thread by any means possible and the recipient cant use modmail bot at all. 
Whats the eval command to close the thread from a recipient?

**Suggestion:** Set a hard limit of how many roles can be displayed in the channel. If it exceed the thresthold, replace it with a phrase like "Too many roles to display."

Perhaps make a new command where it force close all "opened" threads by user. This would be useful feature, specially if theres any future issues that causes channel to not get created by various factors.

PR #200, opened 2019-03-18
Fix for Issue #191

codeinteger6:
Here is the fix for Issue #191. The problem was caused due to Discord's character limit, if the user has too many roles the thread creation process won't work properly. This PR sets a limit on how many roles the bot can display before it says "Too many roles to show" therefore not exceeding Discord's character limit. I have set the value to 21, meaning it can display up to 20 roles before it says too many roles.

Initially I tried setting the value to 41, but then the embed message of where the user info and roles are shown wasn't sending at all, so I reduced the value to 21 and now it's working fine.

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -500,9 +500,9 @@ class Thread(ThreadABC):
 
         await destination.send(mentions, embed=embed)
         if additional_images:
-            self.ready = True
-            await asyncio.gather(*additional_images)
             self.ready = False
+            await asyncio.gather(*additional_images)
+            self.ready = True
 
         if delete_message:
             self.bot.loop.create_task(ignore(message.delete()))
--- a/core/thread.py
+++ b/core/thread.py
@@ -345,7 +345,7 @@ class Thread(ThreadABC):
                                         ))
 
         if self.close_task is not None:
-            # cancel closing if a thread message is sent.
+            # Cancel closing if a thread message is sent.
             await self.cancel_closure()
             tasks.append(
                 self.channel.send(
@@ -436,7 +436,7 @@ class Thread(ThreadABC):
         additional_images = []
         additional_count = 1
 
-        for att in images:  # TODO: Logic needs review
+        for att in images:
             if not prioritize_uploads or (
                     is_image_url(*att) and not
                     embedded_image and
--- a/core/thread.py
+++ b/core/thread.py
@@ -443,6 +443,9 @@ class Thread(ThreadABC):
                     att[1]
             ):
                 embed.set_image(url=att[0])
+                if att[1]:
+                    embed.url = att[0]
+                    embed.title = att[1]
                 embedded_image = True
             elif att[1] is not None:
                 if note:
--- a/core/thread.py
+++ b/core/thread.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 import re
 import string
 import typing
@@ -8,7 +9,11 @@ import discord
 from discord.ext.commands import UserInputError, CommandError
 
 from core.models import Bot, ThreadManagerABC, ThreadABC
-from core.utils import is_image_url, days, match_user_id, truncate, ignore
+from core.utils import is_image_url, days, match_user_id
+from core.utils import truncate, ignore, error
+
+
+logger = logging.getLogger('Modmail')
 
 
 class Thread(ThreadABC):
@@ -313,8 +318,8 @@ class Thread(ThreadABC):
                             destination=self.recipient,
                             from_mod=True,
                             anonymous=anonymous)
-        except Exception as e:
-            print(e)
+        except Exception:
+            logger.info(error('Message delivery failed:'), exc_info=True)
             tasks.append(message.channel.send(
                 embed=discord.Embed(
                     color=discord.Color.red(),
--- a/core/thread.py
+++ b/core/thread.py
@@ -365,13 +365,13 @@ class Thread(ThreadABC):
                 )))
             self.bot.loop.create_task(tasks)
 
+        if not self.ready:
+            await self.wait_until_ready()
+
         if not from_mod and not note:
             self.bot.loop.create_task(
                 self.bot.api.append_log(message, self.channel.id)
             )
-
-        if not self.ready:
-            await self.wait_until_ready()
 
         destination = destination or self.channel
 
@@ -556,6 +556,7 @@ class ThreadManager(ThreadManagerABC):
             if channel:
                 thread = Thread(self, recipient, channel)
                 self.cache[recipient.id] = thread
+                thread.ready = True
         return thread
 
     async def _find_from_channel(self, channel):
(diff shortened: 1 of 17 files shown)


codeinteger6:
It is similar to what Dyno does when using whois on a user with a lot of roles.

fourjr:
i dont think that is relevant to modmail?

codeinteger6:
It will at least solve the issue, so best to deploy this now, if needed changes can be made later

AmitGamingKing:
For your question @fourjr if any user has more roles, then just change value in `if` condition as of now it is set to 41 roles including the `@everyone` role you can change it to a specific limit but as we tried it didn't work for more roles and same problem was happening so this is a basically a quick fix! However devs will surely work on this thing in future commits...

fourjr:
I suggest checking the size of the description and ensure it does not cross the limit.

kyb3r:
Instead of using a role number limit it would be more robust to use a character limit.

codeinteger6:
Will see what can be done and report back.

codeinteger6:
After doing some extended testing and modifications, here is the final version of the thread.py file.

Some things to point out:

1. In non-separate server mode, the amount of characters in the Discord role name does not matter as it uses IDs to take up the character limit. I tried setting the value to 46, but it was exceeding the 1024 character limit. 41 is fine though (meaning 40 roles will be displayed excluding everyone role).

2. In separate server mode, this is where character count is important unlike the non separate mode where it uses mentions so only IDs take up the character count. The bot can handle up to 1024 characters in a single embed. But at the same time, need to make space for other characters in the embed so I decided to allow up to 800 characters (including roles, commas and spaces in total) that can be displayed, so there is enough room for the other characters in the embed, including the full username which can go up to 32 characters.

3. A role count is also added.

This is the best that could be done, hope this is sufficient.

fourjr:
Wouldn't a simple `value[:1000] + '...'` work?

codeinteger6:
Non separate server mode the character count does not matter. Because it counts the Discord IDs only in the character count not the role name as it uses mentions. It can display up to 40 without problems, 45 roles exceeds that limit.

For separate server mode it's set to up to 800 characters for the role list, it needs space for other things like name, join date, number of threads, ID and time in the embed as well.

But I believe that the thread.py file currently is ready to be merged. If necessary, changes can be made later but it will at least not cause problems for users with so many roles in certain servers. I tried to set the limits as high as it can while taking everything into account.

codeinteger6:
I need to check for separate server mode again, don't merge now. I will come back with an updated file if needed.

codeinteger6:
You were right. I just checked the Discord developer docs and the field.value, which is where all the roles are displayed can't exceed more than 1024 characters. I edited the file for both separate and non-separare server modes and now I'm fully confident it is ready to merge and deploy the update.

core/thread.py
```diff
@@ -645,15 +645,30 @@ def _format_info_embed(self, user, log_url, log_count, color):
...
+                if len(charCounter) <= 1024:
+                    role_names = ', '.join(r.name for r in roles
+                                            if r.name != "@everyone")
+                else:
+                    role_names = "Due to Discord limitations the bot can't display roles that contain more than 1024 characters in total."
             else:
-                role_names = ' '.join(r.mention for r in roles
-                                      if r.name != "@everyone")
+                for r in roles:
+                    count = count + 1
```
fourjr: count += 1

core/thread.py
```diff
@@ -680,7 +695,8 @@ def _format_info_embed(self, user, log_url, log_count, color):
                                 value=member.nick,
                                 inline=True)
             if role_names:
-                embed.add_field(name='Roles',
+                count = count - 1
```
fourjr: count -= 1

core/thread.py
```diff
@@ -645,15 +645,30 @@ def _format_info_embed(self, user, log_url, log_count, color):
...
+        count = 0
         if member:
             separate_server = self.bot.guild != self.bot.modmail_guild
             roles = sorted(member.roles, key=lambda c: c.position)
             if separate_server:
-                role_names = ', '.join(r.name for r in roles
-                                       if r.name != "@everyone")
+                for r in roles:
+                    count = count + 1
+                charCounter = ', '.join(r.name for r in roles
```
fourjr: use snake_case

core/thread.py
```diff
@@ -645,15 +645,30 @@ def _format_info_embed(self, user, log_url, log_count, color):
...
+                                            if r.name != "@everyone")
+                if len(charCounter) <= 1024:
+                    role_names = ', '.join(r.name for r in roles
+                                            if r.name != "@everyone")
+                else:
+                    role_names = "Due to Discord limitations the bot can't display roles that contain more than 1024 characters in total."
```
fourjr: i think it's best to show what we can & put ellipsis after

kyb3r:
> Wouldn't a simple `value[:1000] + ('' if len(value) <= 1000 else '...')` work?

This would mess up role mentions, I think it'd be better to iteratively concatenate role names/mentions until the length of the string is just under 1024.

e.g. 
```py
roles = []

for i, r in enumerate(user.roles):
    fmt = r.name if sep_server else r.mention
    if sum(len(r) for r in roles) + len(fmt) > 1000:
        roles.append('...')
        break
    roles.append(fmt)

seperator = ', ' if sep_server else ' '
roles = seperator.join(roles)
```

fourjr:
What about commands.Paginator? It implements that iirc

AmitGamingKing:
@fourjr My opinion is you create a new pull request by your own implementation and show that to us would be great for you & everyone :D

codeinteger6:
What else can be done differently? My [latest commit](https://github.com/kyb3r/modmail/pull/200/commits/9fb31e39b220788904dcec949c9c05dc28f33bd6) I have tested it in both separate and non-separate mode and I believe it is fine the way it is. The problem was the field.value where the role lists are displayed can only go up to 1024 characters according to Discord developer docs. So i made it only display up to 1024 characters if it goes more than that it will say can't show roles but it will at least not crash and cause problems for some servers.

Test it and see if anything needs to be changed but I'm pretty sure there's nothing else that need to be done and ready for deployment.

codeinteger6:
And before starting to edit the thread.py file I checked what the original problem was by reproducing the bug and seeing what comes up in the logs:

```
2019-03-19T08:39:16.782389+00:00 app[worker.1]: future: <Task finished coro=<Thread.setup() done, defined at /app/core/thread.py:79> exception=HTTPException('BAD REQUEST (status code: 400): Invalid Form Body\nIn embed.fields.0.value: Must be 1024 or fewer in length.')>
2019-03-19T08:39:16.782418+00:00 app[worker.1]: In embed.fields.0.value: Must be 1024 or fewer in length.
```
You can see it was the field.value exceeding 1024 characters.

fourjr:
We don't develop modmail just for "it to work" and getting the fastest solution. It is also for style and the cleanest possible code. If you feel that you want to stick to your messy code, feel free to do so in your own fork.

codeinteger6:
Here's a cleaner code

kyb3r:
A bit late but I implemented this in 2db8b854caa14aa2f7434a28fa636a865445c0e3

PR #201, opened 2019-03-19
Update config.json.example

Mamas159:


codeinteger6:
First of all you're not supposed to PR your config.json into here. Second you should revoke that bot token before somebody uses it.

fourjr:
@kyb3r delete this.

PR #209, opened 2019-04-07
Changed lock to emoji

papiersnipper:
afbeelding
This will change 'lock' to 'emoji', which makes more sense in my opinion because the user can set it to any emoji they want.


kyb3r:
The users can change the footer text via editing the `thread_creation_footer` config variable, so this is unnecessary.

PR #214, opened 2019-04-19
invite command

WebKide:
self explanatory


fourjr:
Not necessary in my opinion. Modmail should have (little or) no commands that normal users are expected to execute.

PR #217, opened 2019-04-29
v2.18.0 - new permission system

### LINKED ISSUE #165: Inconsistency with commands structure
So the intention with who can use the mod mail commands to interact with the bot are supposed to be for those who only have manage delete however theres inconsistency with that.

My trial moderators dont have manage message permission to prevent rouge actions but for some reason they do have the ability to use
?reply
?close
?snippets add

I would like to suggest a more clear defined permissions on who can use the command based on Usergroup Roles rather than who has access manage message permission.

### LINKED ISSUE #196: Suggestion: Restructure the permission tree on commands usage.
There are permission inconsistency for different mod mail commands. Taku pointed out that you need manage channel permission to use ?contact command
while you only need manage message permission to use subcommands of alias/snippets and more.

The permission are also undocumented from wikis which can lead to confusion of who can use what commands or not. Because as above mention only specific commands works for specific permission granted for each role. 

Suggestion: Rather then having mod mail check what specific permission a role has, let us add role ID as it adds far more flexibility and let us fully customize the commands permission where we can choose what commands role id can use.

We can also have presets of command permission based on 3 profiles. 
Admin, moderator and helper.
These 3 have already have assigned to specific amount of commands of what they are allowed to use incase user dont want to individually enable specific commands, one at the time.

Taaku18:
Resolves #196 and resolves #165

# v2.18.0

### What's new?

- A brand new permission system! Replacing the old guild-based permissions (ie. manage channels, manage messages), the new system enables you to customize your desired permission level specific to a command or a group of commands for a role or user.
- There are five permission groups/levels:
  - Owner
  - Administrator
  - Moderator
  - Supporter
  - Regular

You may add a role or user to a permission group through any of the following methods:
- `?permissions add level owner @role`
- `?permissions add level supporter member-name`
- `?permissions add level moderator everyone`
- `?permissions add level moderator @member#1234`
- `?permissions add level administrator 78912384930291853`

The same applies to individual commands permissions:
- `?permissions add command-name @member#1234`
- ... and the other methods listed above.

To revoke permission, use `remove` instead of `add`.

To view all roles and users with permission for a permission group or command do:
-  `?permissions get command command-name`
-  `?permissions get level owner`

### Note

When updating to this version, all prior permission settings with guild-based permissions will be invalidated. You will need to convert to the above system.

Check the changelog for the most up-to-date changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+# v2.17.1
+
+### Changed
+
+- Stricter fallback genesis embed search. This update shouldn't affect anyone.
+
+### Info
+How modmail checks if a channel is a thread: 
+
+1. First the bot checks if the channel topic is in the format `User ID: xxxx`, this means it is a thread.
+2. If a channel topic is not found, the bot searches through message history of a channel to find the thread creation embed. This step should never yield a thread for a normal user, but in the case of a another bot messing up the channel topic (happened to a user before) this extra step was added. 
+
+
 # v2.17.0
 
 ### What's new?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# v2.16.1
+
+### Fixed
+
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,46 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+-  `?permissions get command command-name`
+-  `?permissions get level owner`
+
+By default, all newly set up Modmail will have `OWNER` set to the owner of the bot, and `EGULAR` set to @everyone.
+
+The help message no longer conceal inaccessible commands due to check failures.
+
+A `?delete` command, which is an alternative to manually deleting a message. This command is created to no longer require manage messages permission to recall thread messages.
+
+### Note
```
fourjr: If I understand this correctly, this is a **breaking change**. This section should be moved above `What's New` to garner more attention.
```suggestion
### Breaking
```

PR #219, opened 2019-05-01
Remove reliability on Modmail API

fourjr:


fourjr:
@kyb3r @Taaku18, this will be merged in 24h if its ok to go

kyb3r:
Can we keep the metadata posting loop, it doesn't matter if it works or not.

PR #238, opened 2019-05-19
Updated the alias shortcut

immortality223yt:
cuz kyb3r said I can


kyb3r:
What you did was incorrect, correct usage is `alias add r reply` or `alias add alias-name command`

PR #239, opened 2019-05-19
Fixed again

immortality223yt:


cogs/utility.py
```diff
@@ -680,7 +680,7 @@ def cog_unload(self):
         - `{prefix}config set varname value here`
 
         To remove a configuration variable:
-        - `{prefix}config set remove varname`
+        - `{prefix}config remove remove varname`
```
kyb3r: its meant to be `config remove varname` not `config remove remove varname`

PR #242, opened 2019-05-21
Formatting

Agent-Fennec:


Agent-Fennec:
I will probably do comments at some point I just did not have time to do that.

kyb3r:
I'm thinking of using [black](https://github.com/python/black) for formatting in the future.

PR #285, opened 2019-06-10
Added some more

MiTonder:


MiTonder:
Made some changed to Sponsors.json so that it looks Better, making more when i have time

PR #293, opened 2019-06-19
Remove update functionality from the bot.

kyb3r:
This is in favour of the pull app.


kyb3r:
Thoughts?

StephenDaDev:
Yes, we have told people to "not bother" adding GitHub tokens, for weeks now. No real reason to keep it at all. Would this update go along with the No Database in v3?

kyb3r:
I think no db is a bit far off, so i'll merge this now

PR #302, opened 2019-06-23
This introduces `thread_auto_close` and `thread_auto_close_response`.…

### LINKED ISSUE #161: suggestion: auto close ticket after specific time
Most companys support ticket system including discord close the ticket automatically if the user doesnt respond **after** moderator sent a reply within  a specific timeframe normally within 48 hours. 

And allow us to edit a custom message for bot to  send to the recipient explaining why the ticket got closed.
Perhaps allow modmail to just that aswell?

daegontaven:
… Both these can be used to configure modmail to auto close a thread after some time has passed.

`thread_auto_close_response` has a `%t` variable that can be used to insert human friendly time into the closing message.

This does not auto close threads that already exist and will have to be manually closed. But once the variable is configured, it will take it from there for every new thread.

**Take a look at this P.R with two sets of eyes as this is my first attempt at calling methods that I did not write.**

Resolves #161

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -1,15 +1,17 @@
 import asyncio
 import logging
-import re
 import os
+import re
 import string
 import typing
-from types import SimpleNamespace as param
 from datetime import datetime, timedelta
+from types import SimpleNamespace as param
 
 import discord
+import isodate
 from discord.ext.commands import MissingRequiredArgument, CommandError
 
+from core.time import human_timedelta
 from core.utils import is_image_url, days, match_user_id
 from core.utils import truncate, ignore, error
 
@@ -39,6 +41,7 @@ class Thread:
         self.genesis_message = None
         self._ready_event = asyncio.Event()
         self._close_task = None
+        self._auto_close_task = None
 
     def __repr__(self):
         return (
@@ -218,6 +221,10 @@ class Thread:
 
         await self.cancel_closure()
 
+        # Cancel auto closing the thread if closed by any means.
+        if self._auto_close_task:
+            self._auto_close_task.cancel()
+
         if str(self.id) in self.bot.config.subscriptions:
             del self.bot.config.subscriptions[str(self.id)]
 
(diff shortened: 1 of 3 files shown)


core/thread.py
```diff
@@ -336,6 +343,72 @@ def _close_after(self, closer, silent, delete_channel, message):
                 if str(message_id) == str(embed.author.url).split("/")[-1]:
                     return msg
 
+    async def _grab_timeout(
+            self
+    ) -> typing.Union[None, isodate.duration.Duration, timedelta]:
```
fourjr: do this instead in one line:
```python
async def _grab_timeout(self) -> typing.Union[None, isodate.duration.Duration, timedelta]:
```

Also, rename the function to `_fetch_timeout` 😛

core/thread.py
```diff
@@ -336,6 +343,72 @@ def _close_after(self, closer, silent, delete_channel, message):
...
+        seconds = timeout.total_seconds()
+        # seconds = 20  # Uncomment to debug with just 20 seconds
+        reset_time = datetime.utcnow() + timedelta(seconds=seconds)
+        human_time = human_timedelta(dt=reset_time)
+
+        # Grab message
+        close_message = self.bot.config.get(
+            "thread_auto_close_response",
+            f"This thread has been closed automatically after no response from"
+            f" you for {human_time}."
```
fourjr: Use this message instead `This thread has been closed automatically due to inactivity after {human_time}`

core/thread.py
daegontaven: I was following PEP8. Are we using a different code style? 😅

core/thread.py
daegontaven: I'll push that commit when I get back home. Thanks!

core/thread.py
fourjr: We follow our code style lul

whatever looks/works best i guess..

kyb3r:
Just use black for code formatting.

core/thread.py
kyb3r: Just use black for code formatting.

kyb3r:
Can you also reflect the changes in the changelog? Thanks :)

core/thread.py
daegontaven: @fourjr I committed that into one line, but the second black formatting commit reverted it back.

PR #304, opened 2019-06-24
Remove update functionality and github commands.

kyb3r:
This is in favour of the pull app.


kyb3r:
I'll increment to v3 after this since technically this is a breaking change.

PR #332, opened 2019-07-27
Update registry.json

matrix2113:
I made an uptime command but it’s kinda useless. Papier helped me fix it.


fourjr:
What's the propose of this plugin? As you mentioned, it is already possible for the same functionality to be done via the `about` command.

matrix2113:
@fourjr so you only need to see the uptime instead of everything else (ping, etc links)

fourjr:
@kyb3r im voting a no

kyb3r:
There's no use for this plugin

PR #339, opened 2019-08-01
Change Ping to only have 2 numbers after decimal

StephenDaDev:
Instead of ``##.####`` it will only have ``##.##``

### DIFF
--- a/cogs/utility.py
+++ b/cogs/utility.py
@@ -610,7 +610,7 @@ class Utility(commands.Cog):
         """Pong! Returns your websocket latency."""
         embed = Embed(
             title="Pong! Websocket Latency:",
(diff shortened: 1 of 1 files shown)


kyb3r:
but why?

StephenDaDev:
[)](https://api.gh-polls.com/poll/01DH9MN8AWYJC3MDK2DNMK2QMW/Make%20it%202%20after%20the%20decimal%20(Merge%20this)/vote)
[)](https://api.gh-polls.com/poll/01DH9MN8AWYJC3MDK2DNMK2QMW/Leave%20it%204%20after%20the%20decimal%20(Close%20this)/vote)
*The Poll was Taki's Idea*

Taaku18:
From the poll results, it’s determined that this change is unnecessary.

PR #349, opened 2019-08-11
Added Moderator to blocking reason

LasseNiermann:
Whenever a person blocks a user nobody knew *who* did that. 
With this little thing, the moderators' name will be added to the end of the reason.

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -1048,7 +1048,7 @@ class Modmail(commands.Cog):
         ):
             if str(user.id) in self.bot.blocked_users:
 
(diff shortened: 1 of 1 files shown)


cogs/modmail.py
```diff
@@ -1021,18 +1023,18 @@ def format_log_embeds(self, logs, avatar_url):
             return await ctx.send(embed=embed)
 
         if after is not None:
-            reason = after.arg
+            reason = after.arg + f" by {moderator}"
```
fourjr:
```suggestion
            reason = f"{after.arg} by {moderator}"
```

LasseNiermann:
@fourjr Would you please check if I resolved your issue?

PR #353, opened 2019-08-20
Update registry.json

DAzVise:
Server Stats plugin such as member count and other voice channels!


codeinteger6:
@DAzVise seems like the pull bot reset your fork and so nothing was added to the registry

PR #356, opened 2019-08-23
Closes #344

### LINKED ISSUE #344: Suggestion: Show Message ID
It would be nice to have the message ID of the message the user sends modmail.  We couldn't report some mail to Discord since they don't take bot screenshots, but they do take the message ID of the message that was sent in the DM.

Weirdllamas:
image

Closes issue #344 by replacing the footer with the message ID of the DM sent to the modmail

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -1012,6 +1012,8 @@ class Modmail(commands.Cog):
 
         mention = getattr(user, "mention", f"`{user.id}`")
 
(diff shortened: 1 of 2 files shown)


StephenDaDev:
That will do the message ID in the Modmail DM Correct?

Weirdllamas:
> That will do the message ID in the Modmail DM Correct?

Yep! I tested the IDs

Issue #298, opened 2019-06-21
Thread cooldown


Forcellrus:
Add a functionality that allows us to set a cooldown timer, before users are able to open another thread for a certain period of time after closure. For example, users open a thread, we deal with them, close the thread and they message us again, opening a new thread and pinging staff just to write "ty". It's becoming really annoying and we think that adding an option to set a 15 min (configurable time) cooldown between closing and creating a new thread would be of great help. It would also really help with spammers.

StephenDaDev:
Good Feature

StephenDaDev:
This is important, it helps with anti-raid and anti-accidental-creation.

TheKinG2149:
GitHub Actions uwu

StephenDaDev:
I know, I hate it.

groovybluedog:
This really couldn't come sooner!

officialpiyush:
Don't we have thread cooldown's now?

StephenDaDev:
Implemented!

RealCyGuy:
should be closed then

StephenDaDev:
@Taaku18 Apparently I never tagged you on this one

RealCyGuy:
*gasp*

PR #362, opened 2019-08-25
Add cooldown config var

papiersnipper:
Implements issue #298.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,20 +13,25 @@ If you are proposing new features, please discuss them with us in the [developme
 ## We Develop with Github
 We use github to host code, to track issues and feature requests, as well as accept pull requests.
 
-## We Use [Github Flow](https://guides.github.com/introduction/flow/index.html), So All Code Changes Happen Through Pull Requests
-Pull requests are the best way to propose changes to the codebase (we use [Github Flow](https://guides.github.com/introduction/flow/index.html)). We actively welcome your pull requests:
+## We Use [Git Flow](https://atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow)
+![Simple Image Of A Git Flow Workflow](https://nvie.com/img/hotfix-branches@2x.png)
+When contributing to this project please make sure you follow this and name your branches appropriately! 
 
-1. Fork the repo and create your branch from `master`.
+## All Code Changes Happen Through Pull Requests
+Make sure you know how Git Flow works before contributing! 
+Pull requests are the best way to propose changes to the codebase. We actively welcome your pull requests:
+
+1. Fork the repo and create your branch from `master` or `development` according to Git Flow.
 2. If you've added code that should be tested, add tests.
 3. If you've changed APIs, update the documentation.
 4. Ensure the test suite passes.
 5. Make sure your code lints.
 6. Issue that pull request!
 
-## Any contributions you make will be under the MIT Software License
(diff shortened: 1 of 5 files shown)


bot.py
```diff
@@ -574,6 +577,18 @@ def main_color(self) -> int:
                 )
                 guild_age = self.config.remove("guild_age")
 
+        if not isinstance(thread_cooldown, isodate.Duration):
+            try:
+                thread_cooldown = isodate.parse_duration(thread_cooldown)
+            except isodate.ISO8601Error:
+                logger.warning(
+                    "The account age limit needs to be a "
```
Taaku18: "The account age limit", copy-paste problem.

bot.py
```diff
@@ -591,6 +606,15 @@ def main_color(self) -> int:
             logger.warning("Error with 'guild_age'.", exc_info=True)
             self.config.remove("guild_age")
 
+        try:
+            last_log = (await self.api.get_user_logs(message.author.id))[-1]
+            cooldown = datetime.strptime(last_log["created_at"].split(".")[0], '%Y-%m-%d %H:%M:%S') + isodate.parse_duration(self.config["thread_cooldown"])
```
Taaku18: `isodate.parse_duration(self.config["thread_cooldown"])` should be replaced with `thread_cooldown`. and parse created_at with `https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat`.

bot.py
```diff
@@ -591,6 +606,15 @@ def main_color(self) -> int:
...
+        except IndexError:
```
Taaku18: Wouldn't that cause name error later on?

bot.py
```diff
@@ -663,6 +687,16 @@ def main_color(self) -> int:
                         self.blocked_users.pop(str(message.author.id))
                 else:
                     logger.debug("User blocked, user %s.", message.author.name)
+        elif cooldown > now:
+            delta = human_timedelta(cooldown)
```
Taaku18: Don't believe `reaction` is defined.

core/config.py
```diff
@@ -5,11 +5,11 @@
 import typing
 from copy import deepcopy
 
-from dotenv import load_dotenv
 import isodate
 
 import discord
 from discord.ext.commands import BadArgument
+from dotenv import load_dotenv
```
Taaku18: `dotenv` should be back where it was, pip8 import grouping.

core/config.py
```diff
@@ -153,7 +154,7 @@ def populate_cache(self) -> dict:
         config_help_json = os.path.join(
             os.path.dirname(os.path.abspath(__file__)), "config_help.json"
         )
-        with open(config_help_json, "r", encoding="utf-8") as f:
+        with open(config_help_json, "r", encoding="utf-8-sig") as f:
```
Taaku18: Why? https://stackoverflow.com/a/2223926/6622817.

bot.py
papiersnipper: Exposed 👀

bot.py
Taaku18: And also, why don't you implement it the same way as account_age and guild_age, although they look a bit messy. The reason for how the others are implemented is to log it in the block reason and only show the "Message not sent" message once in a user's DM. If you want, just leave it and I can convert it later on.

Taaku18:
It seems like because of the pull app enabled on your fork, your changes were overwritten.

PR #368, opened 2019-09-09
Delete Pipfile.lock

biswajit9123:


Taaku18:
Procfile is specific to Heroku and Heroku very well support Pipenv, therefore, why is the point of this pr? For reference, future pr’s without a description will be automatically closed.

PR #377, opened 2019-09-16
Update modmail.py

DAzVise:
now contact command auto-deletes, sends the embed to the thread channel


Taaku18:
PR to development branch please

Issue #338, opened 2019-07-31
Command to see every thread user has responded in


StephenDaDev:
Similar to ``[p]logs closed-by`` have a subcommand that will show every thread the member has sent a reply in.

StephenDaDev:
@Taaku18 This was added in [v3.3.0](https://github.com/kyb3r/modmail/blob/master/CHANGELOG.md#v330)

Close requested.

StephenDaDev:
oops
i can close it myself

PR #381, opened 2019-09-29
Add ?logs responded command (#338)

papiersnipper:
I also formatted with black and the other files got changed, so I decided to leave them in this commit.


papiersnipper:
I forgot we're having this development branch now...

PR #402, opened 2019-10-29
Delete notes and "?logs delete" command

### LINKED ISSUE #210: Ability to delete certain logs via the bot itself
Suppose you don’t want to keep unnecessary past threads in the database such as test threads and it increases that person’s past thread count. Currently if we want to delete a log from the database we have to log into MongoDB atlas, go to Collections, find it in a long list of logs and delete it. So it would be great if there is a command such as `logdelete` where you input the log URL or the log ID and it deletes it from the database. Also have different ways to use the command, such as deleting an individual item, deleting all logs for a selected user via mention or User ID and a delete everything option.

### LINKED ISSUE #323: Ability to delete notes
Right now you can't delete notes with ``[p]del [ID]``

papiersnipper:
This PR introduces a feature that's already in the `?edit` command. Support members can now delete notes if they provide the message ID. Usage: `?delete [NOTE_ID]`. 
Closes #323.

Secondly, there's a command to delete log entries. Usage: `?logs delete [LOG_KEY_OR_LINK]`.
Closes #210.

As a "QoL improvement", I added a few lines of code to support already existing aliases when creating new ones.


StephenDaDev:
*Changelog?*

Taaku18:
How about make it delete the last message or note, whichever is the latest, rather than requiring to provide the ID.

Issue #303, opened 2019-06-23
[Suggestion] Add an alert on Category filled


Coolguy3289:
The discord limitation of 50 channels per category is well known. I am proposing a chat alert in the logging channel to state that the limit has been hit so the moderators of a server can take proper action to keep tickets flowing, otherwise checking the logs manually is the only way to know what has happened.

maidos:
we need a contingency plan for this because most of us rarely check logs. So we have no way of knowing when we reach 50 channels. Sometimes the recipient arent even aware that the bot couldnt create a thread and assume it was created.

Im suggesting that the bot creates a temporary category once the original category reach 50 channels. And when the original category have fewer than 50 channels, the channels in the temporary category get auto moved to the original category and 2nd category gets deleted

Coolguy3289:
Honestly that feature would be *very* helpful, better than manually dragging channels to an overflow category. I may see about making the changes myself depending on the proposed timeline for release

StephenDaDev:
> most of us rarely check logs

In most larger servers I'm in almost all of the staff mute the log channel. If you absolutely had to keep this concept and couldn't change maybe make it do a mention such as `overflow-mention` also if you would do that maybe make a config var `overflow-warning-channel` so you could send it to Staff Announcements or somewhere staff do actually look

Coolguy3289:
Like I said, the suggested addition of auto creating overflow categories and using that for new tickets to prevent the tickets from being created would be helpful and imo future proofing

Taaku18:
@DAzVise Is working on this.

StephenDaDev:
#636

groovybluedog:
Great to see this is coming, I also support the idea of the bot creating a secondary category and putting threads back in to the main one and then deleting itself when space becomes available.

StephenDaDev:
Overflow category was implemented! View the config help below:
---
Default:
``Fallback Modmail`` (created when the main category is full)

Information:
This is the category that will hold the threads when the main category is full.

To change the Fallback category, you will need to find the category’s ID.

Example(s):
- ``?config set fallback_category_id 9234932582312`` (9234932582312 is the category ID)
Note(s):
- If the Fallback category ended up being non-existent/invalid, Modmail will create a new one. To fix this, set ``fallback_category_id`` to a valid category.
- See also: ``main_category_id``.
---
@Taaku18 *implemented*

PR #636, opened 2019-11-15
added fallback category. #303

DAzVise:


core/thread.py
```diff
@@ -859,11 +859,25 @@ def create(
         self.cache[recipient.id] = thread
 
         # Schedule thread setup for later
+        cat = self.bot.main_category
+        if len(cat.channels) == 50:
+            fallback_id = self.bot.config["fallback_category_id"]
+            fallback = discord.utils.get(cat.guild.categories, id=int(fallback_id))
```
Taaku18: As `fallback_id` is `None` by default, line 865 will error. You'll need to add a `if fallback_id is not None` check from line 685 - 688:

    if fallback_id is not None:
        fallback = discord.utils.get(cat.guild.categories, id=int(fallback_id))
        if fallback and len(fallback.channels) != 50:
            self.bot.loop.create_task(thread.setup(creator=creator, category=fallback))
            return thread

core/thread.py
```diff
@@ -859,11 +859,25 @@ def create(
...
+                return thread
+
+            fallback = await cat.clone(name="Fallback Modmail")
+            self.bot.config.set("fallback_category_id", fallback.id)
+            await self.bot.config.update()
+            self.bot.loop.create_task(thread.setup(creator=creator, category=fallback))
+            return thread
+
         self.bot.loop.create_task(thread.setup(creator=creator, category=category))
         return thread
```
Taaku18: Instead of keeping three references for creating the thread (line 687, 874, 876), use `if`s and redefine `cat` to the new category's ID, and keep only line 876 for creating the new thread.

core/thread.py
DAzVise: Nope, it will not error, would just return None, which is the reason I check `if fallback` in line 866

officialpiyush: It will still raise an error as fallback channel config variable has a `None` value. And if fallback channel variable isn't set,  `int(None)`  would anytime raise an error 😉

DAzVise: o ya 😅

core/thread.py
DAzVise: Great Idea! done.

PR #638, opened 2019-11-17
Add rules plugin to registry

StephenDaDev:
The title says all of it but, Hello.

### DIFF
--- a/plugins/registry.json
+++ b/plugins/registry.json
@@ -178,5 +178,14 @@
         "title": "Server Stats",
         "icon_url": "https://i.gyazo.com/fadb70740e83f2448b23ffe192a1f32d.png",
         "thumbnail_url": "https://i.gyazo.com/fadb70740e83f2448b23ffe192a1f32d.png" 
+    },
+     "rules": {
+        "repository": "StephenDaDev/modmail-plugins",
+        "branch": "master",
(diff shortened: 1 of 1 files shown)


officialpiyush:
Oh oh oh StephenDaDev into plugins woooooooooooooooo

StephenDaDev:
lol

DAzVise:
use cases? i think modmail's use case is for big servers which probably already have a set of rules.

officialpiyush:
Seems like tough competition ahead 🕵️‍♀️

officialpiyush:
@DazVise i guess you forgot, modmail is used by servers having > 10 members too 😂🙄

DAzVise:
@officialpiyush so do you think this plugin should get added to the plugins registry? not to be rudebut it's just text which should not be part of a bot's functionality

StephenDaDev:
Some people, the first thing they do is deploy a Modmail, this could also assist with helping them get their server ready faster, and will just serve a starter for them to change their rules as needed, (Also for lazy ppl like piyush)

StephenDaDev:
(that sounded kinda mean, I wasn't trying to insult piyush)

officialpiyush:
Yes I'm lazy kiddo 😁🤧

Akhil2149:
Not fancy. Embedder plugin or announcement plugin can already do this.

PR #651, opened 2019-11-27
Development v3.2.2

Taaku18:
# v3.3.2-dev4

### Added

- Thread cooldown!
  - Set via the new config var `thread_cooldown`.
  - Specify a time for the recipient to wait before allowed to create another thread.
- Fallback Category (thanks to DAzVise PR#636)
  - Automatically created upon reaching the 50 channels limit.
  - Manually set fallback category with the config var `fallback_category_id`.
- "enable" and "disable" support for yes or no config vars.
- Added "perhaps you meant" section to `?config help`.
- Multi-command alias is now more stable. With support for a single quote escape `\"`.
- New command `?freply`, which behaves exactly like `?reply` with the addition that you can substitute `{channel}`, `{recipient}`, and `{author}` to be their respective values. 

### Changed

- The look of alias and snippet when previewing.

### Fixed

- Setting config vars using human time wasn't working.
- Fixed some bugs with aliases.

### Internal

- Commit to black format line width max = 99, consistent with pylint.
- Alias parser is rewritten without shlex.


codeinteger6:
We are on v3.4.1 now

PR #2748, opened 2019-12-04
Fix the contact command if more than one arguments are passed

officialpiyush:
Fixes: ->
image
and 
image


officialpiyush:
nvm
image

PR #2749, opened 2019-12-08
Add giveaway plugin

officialpiyush:
( works )


officialpiyush:
why tf does it conflict

officialpiyush:
Fixed da conflicts boizzz

PR #2751, opened 2019-12-14
Update v3.4.0

Taaku18:


Taaku18:
While translation's being worked on in the Dev branch, final testings will be done here. If there are no problems, we will merge is branch in 24 hours. If you found a bug or an issue on this branch, please mention below.

PR #2767, opened 2020-02-21
Feature: Support private repositories for plugins

officialpiyush:
# Use cases:

* When someone doesn't want their plugin to be public, etc.

I have tested this and use it in 3 of my production bots, still, I would like if the community tests this too. Cheers!


officialpiyush:
@Taaku18 @fourjr Bump

RealCyGuy:
we should get this

PR #2777, opened 2020-03-14
Update registry.json

davidm5375:


officialpiyush:
Hey,
There are two things you should fix:

1. The plugin won't be loaded ( would error while installation ) because you have not defined a setup function, plus you need to remove these lines - https://github.com/davidm5375/Musiq-for-Modmail/blob/cd558d65804ad28a81d3b28c0961027f118ed6e5/musiq-for-modmail/musiq.py#L493-L494 and replace it with setup function. 

2. The folder name and the file name must be the same.
folder-struct

You can read more here - https://github.com/kyb3r/modmail/wiki/Plugins

Taaku18:
Does this work? Can anyone confirm its validity?

mischievousdev:
No @Taaku18 its not working, tested!

aboutDavid:
Yeah you can close it if you want.

PR #2784, opened 2020-04-16
Ensure optional dependencies in pyproject.toml can be installed by poetry.

daegontaven:
Right now trying to install `motor` fails if you try it with poetry.

This adds an extras section so that the defined optionals can be installed with poetry.
Source: https://stackoverflow.com/a/60990574/5586359

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,3 +54,5 @@ black = {version = "=19.3b0", allows-prereleases = true}
 pylint = "^2.4"
 bandit = "^1.6"
 
+[tool.poetry.extras]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -54,3 +54,5 @@ black = {version = "=19.3b0", allows-prereleases = true}
 pylint = "^2.4"
 bandit = "^1.6"
 
+[tool.poetry.extras]
+mongodb = ["motor"]
```
daegontaven, line 58: It's called `mongodb` assuming dependencies apart from `motor` will be required in the future.

PR #2786, opened 2020-04-20
make approved issues not stale

RealCyGuy:


StephenDaDev:
What do you think about adding the labels "bug" as well as "maybe:bug"

RealCyGuy:
maybe not maybe: bug

StephenDaDev:
Yeah, I'd have to agree with that within 100 days, it should be decided

Coolguy3289:
Any reason why this hasn't been merged yet?

PR #2787, opened 2020-04-30
Python3.8

ghost:


requirements.min.txt
```diff
@@ -2,23 +2,22 @@
...
+-i https://pypi.org/simple
+aiohttp==3.6.2
 async-timeout==3.0.1
 attrs==19.3.0
 chardet==3.0.4
-discord.py==1.2.5
 dnspython==1.16.0
 emoji==0.5.4
-future==0.18.1
+git+https://github.com/Rapptz/discord.py.git@c62b6c3e8815cec6f73ec20aa4d0ebd9b68a0f14#egg=discord-py
```
daegontaven, line 12: Why is this using the git archive instead of a direct pinning of the package from pypi?

Taaku18:
will re-pr later

PR #2788, opened 2020-04-30
Development

MrMonsterKong:


core/paginator.py
```diff
@@ -176,7 +176,7 @@ async def close(self, delete: bool = True) -> typing.Optional[Message]:
         """
         self.running = False
 
-        sent_emoji, _ = await self.ctx.bot.retrieve_emoji()
+        sent_emoji, x = await self.ctx.bot.retrieve_emoji()
```
SpyTec, line 179: Is there a reason for changing to `x`? Underscore usually denotes the return is throwaway and not used

PR #2789, opened 2020-04-30
Feature(plugins) support private repositories

ghost:


Coolguy3289:
Is there a reason this PR was created when one already exists for the same feature?

PR #2791, opened 2020-05-09
Modmail Thread Invite detection

Lucifer-Mornings:
I suggest a new feature added to the core of modmail, Server Invite Detection. it should check if the user = Supporter (like server staff) or if it's just a regular user, and if the user is not supporter, it removes the invite Link. this should be set to work for both Modmail threads, and server channels together. Sort of like a anti-advertisement tool.

See BASIC example code here;

 if (message.content.includes("discord.gg/")) {
    message.reply("dont advertise"); 
    message:delete();
  }
});


chrissie1234:
I suggest someone to make a plugin for this, so it won't affect the core functionality and people can opt-in by installing the plugin.

Taaku18:
This is a GH Issue, not a PR.

PR #2802, opened 2020-06-01
Added new plugin

Teen1:
added slowmode plugin to registry


Taaku18:
Lmk when this is tested and people can confirm working as intended :)

PR #2813, opened 2020-07-02
Add mongo_db_name into the Config

RheaAyase:
Not tested yet cause we've got a lot more stuff to fix before we can run modmail.

Just going off of [the docs](https://motor.readthedocs.io/en/stable/api-asyncio/asyncio_motor_client.html)

cc @SpyTec

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,3 +3,4 @@ LOG_URL=https://logviewername.herokuapp.com/
 GUILD_ID=1234567890
 OWNERS=Owner1ID,Owner2ID,Owner3ID
 MONGO_URI=mongodb+srv://mongodburi
+MONGO_DB_NAME=modmail_bot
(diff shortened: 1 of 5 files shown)


RheaAyase:
Tested and it's fine.

RheaAyase:
So the logviewer has the database name also hardcoded, so this change in here doesn't make much sense without also changing that thing. My solution is to pass the db name in the [URL itself](https://github.com/RheaAyase/logviewer/commit/912fbd5fdb18cd3f95583e7d8064fc1e571eb737), that way fully supporting multiple instances of modmail on one server, with one instance of logviewer. This would be a breaking change though and I don't know how to make it backwards compatible in the api so I think that this PR leads nowhere.

viharm:
What was the resolution for this? How can we specify database name?

PR #2814, opened 2020-07-02
Allow a custom config dir using cmd line variables

RheaAyase:
Use-case is hosting multiple instances of the bot using a single mongo database (#2813) on a single host. Just pass a path to the config dir for each instance as an argument. Config dir supports both dotenv and config.json

As far as I can tell it works as intended.

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -5,7 +5,6 @@ import re
 import typing
 from copy import deepcopy
 
-from dotenv import load_dotenv
 import isodate
 
 import discord
@@ -17,7 +16,6 @@ from core.time import UserFriendlyTimeSync
 from core.utils import strtobool
 
 logger = getLogger(__name__)
-load_dotenv()
 
 
 class ConfigManager:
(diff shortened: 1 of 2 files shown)


RheaAyase:
I ain't got no clue what `black` is and I can't check it. The CI output is plain useless so unless someone tells me what's wrong, can't fix it.

fourjr:
Hey, running 
```bash
pip install black
black .
``` should autoformat and fix the issues

**Comment by RheaAyase** [comment]
> Hey, running
> 
> ```shell
> pip install black
> black .
> ```
> 
> should autoformat and fix the issues

As I said I can not do that on my current workstation machine and I will not do that on my servers.

File: core/config.py [hunk 1]
```diff
@@ -134,7 +134,7 @@ class ConfigManager:
     defaults = {**public_keys, **private_keys, **protected_keys}
     all_keys = set(defaults.keys())
 
-    def __init__(self, bot, config_path = os.path.dirname(os.path.abspath(__file__))):
+    def __init__(self, bot, config_path=os.path.dirname(os.path.abspath(__file__))):
```
RheaAyase, line 137: So the tool suggests using operators without spaces? That couldn't be more wrong lol :D

RheaAyase:
As I mentioned in the PR #2813 this also seems to lead nowhere without those things. I think that we can close it unless someone has some better ideas how to achieve that (while I'll maintain my own forks with those changes necessary.)

RheaAyase:
The longer I'm running this thing on my server the more I regret ever touching it. Closing PRs as I'm deleting my fork in order to write my own solution that will be reliable and won't spit out silent errors all the time.

PR #2815, opened 2020-07-03
Fixed mention not working with log_url disabled

RheaAyase:
For some reason it was set to mention only if creator didn't exist. Doesn't make any sense...

### DIFF
--- a/app.json
+++ b/app.json
@@ -21,7 +21,7 @@
     },
     "LOG_URL": {
       "description": "The url of the log viewer app for viewing self-hosted logs.",
-      "required": true
+      "required": false
     }
   }
 }
--- a/bot.py
+++ b/bot.py
@@ -741,7 +741,7 @@ class ModmailBot(commands.Bot):
                 await self.add_reaction(message, blocked_emoji)
                 return await message.channel.send(embed=embed)
 
-            thread = await self.threads.create(message.author)
+            thread = await self.threads.create(message.author, creator=message.author)
         else:
             if self.config["dm_disabled"] == 2:
                 embed = discord.Embed(
--- a/core/clients.py
+++ b/core/clients.py
@@ -166,6 +166,7 @@ class ApiClient(RequestClient):
         prefix = self.bot.config["log_url_prefix"].strip("/")
         if prefix == "NONE":
             prefix = ""
+
         return f"{self.bot.config['log_url'].strip('/')}{'/' + prefix if prefix else ''}/{key}"
 
     async def delete_log_entry(self, key: str) -> bool:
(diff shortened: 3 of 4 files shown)


DAzVise:
It does make sense, so when commands like ?contact are ran it shouldn't really ping

RheaAyase:
> It does make sense, so when commands like ?contact are ran it shouldn't really ping

And yet it *should* ping when they're not - it doesn't, ever. `creator` is always there. The code that I changed made no sense and wasn't the solution to the problem that you're mentioning.

RheaAyase:
...which leads me to believe that this code will never run either:
https://github.com/kyb3r/modmail/blob/4953712d8dc00965f69878ad662e993fbf6540a0/core/thread.py#L164-L165

RheaAyase:
Let me just check if I understand this correctly:

* `creator` is supposed to be present only if staff creates the thread. 

* `recipient` is present always and is the PMing user.

* `creator` is not present if `recipient` started the thread.

-> Purely logically ignoring the existing code, shouldn't `creator` be present always cause someone always created it, and then these conditions be `creator == recipient` ? It would be much more readable code, and actually make sense wink wink.

Something like.... like this \*commit follows*

RheaAyase:
...which obviously doesn't actually fix the mention issue because the code path is the same. Looking to figure out why it's executed that way. Please keep holding my beer for a bit longer.

RheaAyase:
While I'm at it, thread and thread manager should be in separate files. It's chaos.

RheaAyase:
Turns out that the problem is somewhere else. This PR needs to be renamed into "fixing absolutely random shit" at this point.

This bot should function without the privacy and security backdoor via web-logs.

lorenzo132:
Heya @RheaAyase 

 may i ask if you can create an entire list of issues you see or any stuff that may be changed and send this in 1 messages, it'll save the ticket from being getting filled out with all small salvos of messages and it'll be easier to read. 

Cheers

RheaAyase:
Issues:

* mongodb
* python

...seriously though? None that I would care about enough to continue fixing. It works as intended, seems that if you use the logviewer then it should work without this PR as well. I do have one beef not mentioned yet - the bot never checks permissions if it can actually do something, for example when you use the ?permissions command it tries to modify access permissions to the category without checking. _(I as admin set that up, not a random bot.)_

This PR can be closed if you think that the creator shouldn't be always present (someone always creates the thread so I think it should.) And if you think that logviewer is absolutely mandatory and one should never use this bot without it.

RheaAyase:
The longer I'm running this thing on my server the more I regret ever touching it. Closing PRs as I'm deleting my fork in order to write my own solution that will be reliable and won't spit out silent errors all the time.

PR #2831, opened 2020-07-23
added translate :wave:

WebKide:
this command conflicts with Translator...


DAzVise:
i think with the addition of this plugin the other plugin should be removed, it doesn't make sense to have 2 translation plugins

sarrafsumit:
Yes

sarrafsumit:
Yes You Are Right

fourjr (approved):

PR #2843, opened 2020-08-27
Update utility.py

onzjz:


cogs/utility.py
```diff
@@ -292,15 +292,13 @@ async def about(self, ctx):
...
-        desc += "members to easily communicate with server administrators in "
-        desc += "an organised manner."
+        desc = "This is an open source Discord bot that has been made by you."
         embed.description = desc
 
         embed.add_field(name="Uptime", value=self.bot.uptime)
         embed.add_field(name="Latency", value=f"{self.bot.latency * 1000:.2f} ms")
         embed.add_field(name="Version", value=f"`{self.bot.version}`")
-        embed.add_field(name="Authors", value="`kyb3r`, `Taki`, `fourjr`")
+        embed.add_field(name="Authors", value="`kyb3r`, `Taki`, `fourjr`, `TheSumit`,")
```
StephenDaDev, line 301: The Authors List is for Core contributors who have made **significant** contributions to the bot, and are considered core developers. Redoing a single description, and removing the word pong, is not in any way, a significant contribution.

cogs/utility.py
```diff
@@ -292,15 +292,13 @@ async def about(self, ctx):
         )
         embed.set_thumbnail(url=self.bot.user.avatar_url)
 
-        desc = "This is an open source Discord bot that serves as a means for "
-        desc += "members to easily communicate with server administrators in "
-        desc += "an organised manner."
+        desc = "This is an open source Discord bot that has been made by you."
```
StephenDaDev, line 295: The about command is made for all members, not just the owner. Claiming it was made by them, is wrong (for almost every user who can use it) and could be confusing on what the actual purpose of the bot is. The description you removed would be more beneficial, as it would explain what the bot was, to any member who used the command.

cogs/utility.py
```diff
@@ -630,7 +628,7 @@ async def before_loop_presence(self):
     async def ping(self, ctx):
         """Pong! Returns your websocket latency."""
         embed = discord.Embed(
-            title="Pong! Websocket Latency:",
+            title="Websocket Latency",
```
RealCyGuy, line 631: I don't think this change is neccesary.

StephenDaDev: Well me either

StephenDaDev:
If you meant to make these on your own bot, and are trying to claim you are an author, stop like seriously.

sumit309:
Het

PR #2845, opened 2020-08-30
Add a "no command description" in the help command.

Vincysuper07:
This should work, for example, when someone tries to see the Moderation plugin. Commands in that plugin have no description, so it will show "No description." instead of nothing.


fourjr:
Your PR should not be changing the code style outside of the main PR reason

Vincysuper07:
I know, but the code-style thing was failing, so I thought to reformat the code...

fourjr:
Run `black` to fix the styling

Vincysuper07:
That's what I did.

fourjr (approved):

PR #2848, opened 2020-09-02
Add run on repl.it badge to README

RobloxStreetMagic:
This pull request adds a  badge to the . This will allow users to easily run this repository in their browser, without having to set up an environment. You can learn more about Repl.it [here](https://repl.it).

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "python3"
+run = ""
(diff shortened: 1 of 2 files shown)


RealCyGuy:
I don't think this is necessary.

fadedmax:
It will work, the readme part can go tho, it doesnt need to be at the very top.
it might be a good edition tho, i just dont think repi.it is very widley used since its public

ririko5834:
Good idea

fourjr:
Hey, officially we do not plan to support repl.it yet due to its disadvantage of "requiring a ping every 30 minutes". We might look at this again in the future though.

PR #2850, opened 2020-09-04
Database

lorenzo132:
Making it a bit easier for people to understand


Forcellrus:
You can't comment in an .env file using // and "actuall" is gramatically incorrect.

PR #2858, opened 2020-10-01
Remove plugins

robinmahieu:
I'm not really able to maintain these anymore, so I don't want people thinking they're safe and approved.


StephenDaDev:
*there goes a decent amount of the ones that actually work*

PR #2864, opened 2020-10-25
Add mediaonly plugin

lorenzo132:
add media-only plugin, 4jr's plugin edited


fourjr (changes requested):
PR is framed in a way such that it says that it only allows images in channels. However, code also does other things such as adding reactions. Misleading description.

In addition, there is no way to set the `config['emojis']` and the bot will probably error out?

lorenzo132:
Hey, i'v removed the emoji functionality's to make sure it will not bug out

fourjr (approved):

fourjr:
Merged. [https://github.com/kyb3r/modmail/commit/7da970c17666b26149fbda4586fa5a79ec8b2c99]

PR #2873, opened 2020-11-04
Replace moderation plugin

Vincysuper07:
(for the commit message, I just accidentally pressed enter)


StephenDaDev:
https://github.com/Vincysuper07/modmail-plugins/blob/main/moderation/moderation.py

StephenDaDev (approved):
The plugin looks fine as well as the edits to the registry

PR #2887, opened 2020-11-15
Added `thread_creation_silent`

zoemartin01:
I just added `thread_auto_open_silently` with the default value 'false' as this allows for the users customisation on a recently changed feature.


DAzVise:
You forgot to add the config in `core/config_help.json`

core/config_help.json
```diff
@@ -303,6 +303,17 @@
       "See also: `thread_creation_title`, `thread_creation_response`, `thread_self_closable_creation_footer`, `thread_close_footer`."
     ]
   },
+  "thread_creation_silent": {
+    "default": "No",
+    "description": "Setting this configuration will open a new thread silently by default.",
```
fourjr, line 308:
```suggestion
    "description": "Setting this configuration will always open a new thread silently in contact.",
```

fourjr (changes requested):
Config variable should be changed to `thread_contact_silent` as it only affects contact

core/config.py
```diff
@@ -53,6 +53,7 @@ class ConfigManager:
         "thread_auto_close_response": "This thread has been closed automatically due to inactivity after {timeout}.",
         "thread_creation_response": "The staff team will get back to you as soon as possible.",
         "thread_creation_footer": "Your message has been sent",
+        "thread_creation_silent": False,
```
fourjr, line 56: it has to be added to `booleans` too

fourjr:
Thanks for the contributions! This PR has been merged in https://github.com/kyb3r/modmail/commit/a123bf617603abb9a1f32f74832595d2d508372b, not sure why the PR wasn't autoclosed.

PR #2927, opened 2021-01-07
Fix `?perms get` command showing IDs instead of user or role mentions.

Jerrie-Aries:
This occurred since bot `v3.6` or `v3.7` if I'm not mistaken, because we changed the user or role IDs (for Permissions Level/command permissions) that we saved to database into `str`.

Before fix:
before

After fix:
after


fourjr:
Thank you!

Issue #2876, opened 2020-11-06
Support editing and deleting plain messages


fourjr:
Related to #2872 

Current limitations:
- Edit and delete use embed author url to find linked messages
- Plain messages do not make use of embeds and do not have this url

Solution:
- Link messages in database

Timeline: Post-v4.0 #2794

wolfier74:
Good idea

wolfier74:
Please add this

wolfier74:
Save message ids or links to database, when thread is closed remove it from database

fourjr:
There is no use spamming our issues with "bumps". If you know want to add the feature yourself, you can learn python and create a pull request.

Jerrie-Aries:
What about if we grab the plain message id in DM channel and put it in `author.url` (or `footer`) when creating an embed for it in thread channel. Since it's just one way plain (in DM), right?
Would it be possible to link those messages later?

Jerrie-Aries:
Now, with multi users support for threads, I think saving the message IDs in the footer is no longer a good idea.

I 'd still suggest using the database for this feature. I have the models for this and been using it on my instance for about a year. It's similar with my last PR (closed) plus some improvements.

I'm still interested in this but don't have time to do it.

lorenzo132:
This has been solved in https://github.com/modmail-dev/Modmail/pull/3416


I'll be closing this issue.

PR #2937, opened 2021-01-14
Support for linking plain message.

Jerrie-Aries:
Support for linking plain message, regarding issue #2876.
As I tested it, delete and edit both worked.

Please check it. Any feedback would be appreciated.

Update:
-Added couple more commits.

And also please let me know if I used wrong methods or if there's anything needed change.

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -947,13 +947,15 @@ class Modmail(commands.Cog):
         thread = ctx.thread
 
         try:
-            await thread.edit_message(message_id, message)
-        except ValueError:
+            await thread.edit_message(ctx.author, message_id, message)
+        except ValueError as e:
+            if "False author." in str(e):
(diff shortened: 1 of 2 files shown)


fourjr:
This PR addresses multiple issues at once:
- Linking to plain messages
- The ability for different mods to edit another mods' message.

Each PR should only address one issue. The ability for different mods to edit messages is an **intended feature**.

If intended features are to be disabled, config options should be added.

fourjr (changes requested):
This PR addresses multiple issues at once:
- Linking to plain messages
- The ability for different mods to edit another mods' message.

Each PR should only address one issue. The ability for different mods to edit messages is an **intended feature**.

If intended features are to be disabled, config options should be added.

Jerrie-Aries:
Ok, sorry for that. Thanks for letting me know 👍. I'll close this.

PR #2945, opened 2021-01-26
Update README.md

Agent-Fennec:
* Update badge to the Latest Version.
* Add description of Heroku and Docker.
* Different hosting options are more clearly marked.


fourjr:
We would appreciate the docker writeup in a separate PR. For now, the other changes will be rejected.

PR #2966, opened 2021-03-07
Use the ID attribute of DM Channel.

scragly:
Originally used the channel instance by mistake, causing the `__str__` to be used instead of the intended ID.


fourjr (approved):
Approved for future release with next version bump

PR #2981, opened 2021-03-23
Register a plugin

elfq:


fourjr:
Rejected, plugin only has one command, should at least have a few various commands

PR #2996, opened 2021-04-08
Added claim plugin to registry.

CoalByte:
Add the claim plugin to the registry file.

Link: https://github.com/CoalByte/better-modmail-plugins/tree/main/claim


fourjr:
Hey, I have made a [claim plugin](https://github.com/fourjr/modmail-plugins/blob/master/claim/claim.py) as well. After discussion with the rest of the team, we have decided to add mine instead.

Sorry about that!

PR #2997, opened 2021-04-09
ADD more option to get the archives (LOGS_URL)

Kathou99:
my SQLite logs will bring the bot at another level and your trust more the direct logs like my sqllite because  he run inside of the bot it's self


Taaku18:
this is not how you update, change the PR direction

PR #3003, opened 2021-04-19
Change Max Channel per Category to 49

Coolguy3289:
Addresses https://github.com/kyb3r/modmail/issues/3002
Will add a 1 channel buffer to the check to account for Discord API bugs preventing the category from actually reaching 50 channels.

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -1205,7 +1205,7 @@ class ThreadManager:
 
         # Schedule thread setup for later
         cat = self.bot.main_category
-        if category is None and len(cat.channels) == 50:
+        if category is None and len(cat.channels) == 49:
             fallback_id = self.bot.config["fallback_category_id"]
             if fallback_id:
                 fallback = discord.utils.get(cat.guild.categories, id=int(fallback_id))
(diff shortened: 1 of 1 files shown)


core/thread.py
```diff
@@ -1205,7 +1205,7 @@ async def create(
 
         # Schedule thread setup for later
         cat = self.bot.main_category
-        if category is None and len(cat.channels) == 50:
+        if category is None and len(cat.channels) == 49:
```
Jerrie-Aries, line 1208:
```suggestion
        if category is None and len(cat.channels) >= 49:
```

Jerrie-Aries (changes requested):

Jerrie-Aries:
Tbh, this PR is not complete yet. I think you need to edit the `if` statement for `fallback` category as well.
In `thread.py` line 1212:
```
                if fallback and len(fallback.channels) <= 49:
```

And also about the solution, since it's the glitch/lag on discord side, I'm unsure whether it should be handled in the code here, or in the `exception` when creating the thread.
So, let the devs look at this first.

fourjr:
This PR has been manually merged.  Thanks for the contribution.

PR #3005, opened 2021-04-25
Change discord.py version to latest (1.7.1)

elfq:


fourjr (commented):
Changing a version is more extensive than simply updating Pipfile. Other requirement files need to be checked as well. The bot needs to be re-tested on the new version keeping in mind any breaking changes.

For now, is there any specific need to have new features from 1.7 in modmail?

PR #3009, opened 2021-04-28
Fix return types, type hints, and unresolved references

Jerrie-Aries:
**Reference:**
- [Message.add_reaction](https://discordpy.readthedocs.io/en/latest/api.html?#discord.Message.add_reaction)


fourjr (approved):
LGTM

PR #3010, opened 2021-04-28
Fix bug on command `?config remove`

Jerrie-Aries:
Fix bug bot doesn't respond on command `?config remove <key>` if invalid `key` is provided due to embed field's value exceeds the limit.

**Steps to reproduce:**
- Run command `?config remove <invalid key here>`.

**Error:**
```py
discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
In embed.fields.0.value: Must be 1024 or fewer in length.
```

### DIFF
--- a/cogs/utility.py
+++ b/cogs/utility.py
@@ -819,7 +819,9 @@ class Utility(commands.Cog):
                 embed = exc.embed
         else:
             embed = discord.Embed(
-                title="Error", color=self.bot.error_color, description=f"{key} is an invalid key."
+                title="Error",
+                color=self.bot.error_color,
+                description=f"`{key}` is an invalid key.",
             )
             valid_keys = [f"`{k}`" for k in sorted(keys)]
             embed.add_field(name="Valid keys", value=truncate(", ".join(valid_keys), 1024))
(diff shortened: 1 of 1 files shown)


Jerrie-Aries:
Umm guys, developers especially! Is it better to use `difflib.get_close_matches()` here?
If it is, I'll close this PR and create another one that uses that method instead, in `?config set`, `?config get`, `?config remove`, `?config help`. So the closest match keys (if any) will be returned (instead of all valid keys) when invalid *key* is provided when running those commands.
And `truncate` probably will no longer be used here.

onerandomusername:
> 
> 
> Umm guys, developers especially! Is it better to use `difflib.get_close_matches()` here?
> If it is, I'll close this PR and create another one that uses that method instead, in `?config set`, `?config get`, `?config remove`, `?config help`. So the closest match keys (if any) will be returned (instead of all valid keys) when invalid _key_ is provided when running those commands.
> And `truncate` probably will no longer be used here.



> 
> 
> Umm guys, developers especially! Is it better to use `difflib.get_close_matches()` here?
> If it is, I'll close this PR and create another one that uses that method instead, in `?config set`, `?config get`, `?config remove`, `?config help`. So the closest match keys (if any) will be returned (instead of all valid keys) when invalid _key_ is provided when running those commands.
> And `truncate` probably will no longer be used here.

Seems like using difflib would make more sense.

To make the change, you don't need to open a new issue. Just make another commit to [your branch for this issue](https://github.com/Jerrie-Aries/modmail/tree/patch-02)

PR #3011, opened 2021-04-29
Update database after resetting/purging all plugins.

Jerrie-Aries:
__About this PR:__
Minor updates on `?plugin reset` command.
- Updates the database after resetting/purging all plugins.
- Removes unloaded plugins from `self.loaded_plugins` which will be shown using `?plugin loaded` command.

*Updated*


Jerrie-Aries:
I just noticed the unloaded plugins were also not removed from `self.loaded_plugins` after purging.
I may push another commit for it.

PR #3012, opened 2021-05-01
Support for linking plain messages.

Jerrie-Aries:
Regarding issue #2876 

__**About this PR:**__
- Support for linking plain messages.
- Deleting and editing plain messages both work.
- Now uses database to store and fetch the linked IDs.
- Add parameter `linked_ids` (type: list) for method `append_log` in `clients.py`. The linked IDs will be stored when the thread message is sent whether from thread or DM.
- No longer find linked messages matching from embed author url.
- Add another return value (type: str) for method `find_linked_messages` in `thread.py`, which would be the type of the thread message. Expected "thread_message", "anonymous", or "system".
- Add parameter `author` (type: [discord.Member](https://discordpy.readthedocs.io/en/latest/api.html#member)) for method `edit_message` in `thread.py`.

__**Note:**__
- Linking multiple images to a single message is still not supported.

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -1286,7 +1286,7 @@ class ModmailBot(commands.Bot):
             if not thread:
                 return
             try:
-                _, linked_message = await thread.find_linked_messages(
+                _, linked_message, _ = await thread.find_linked_messages(
                     message.id, either_direction=True
                 )
             except ValueError as e:
--- a/bot.py
+++ b/bot.py
@@ -1279,7 +1279,8 @@ class ModmailBot(commands.Bot):
                     message, either_direction=True
                 )
             except ValueError as e:
-                logger.warning("Failed to find linked message for reactions: %s", e)
+                if str(e) != "Ignored message.":
+                    logger.warning("Failed to find linked message for reactions: %s", e)
                 return
         else:
             thread = await self.threads.find(channel=channel)
@@ -1290,7 +1291,8 @@ class ModmailBot(commands.Bot):
                     message.id, either_direction=True
                 )
             except ValueError as e:
-                logger.warning("Failed to find linked message for reactions: %s", e)
+                if str(e) != "Ignored message.":
+                    logger.warning("Failed to find linked message for reactions: %s", e)
                 return
 
         if self.config["transfer_reactions"] and linked_message is not None:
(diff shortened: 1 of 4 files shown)


Jerrie-Aries:
By the way, this still can be improved.
For example remove the mod/recipient embed color and the message type (Note, Persistent Note) checks, and do those checks using the data we get from the database instead.

So the example of code for `find_linked_messages` in `thread.py` would be:
```py
    async def find_linked_messages(
        self,
        message_id: Optional[int] = None,
        either_direction: bool = False,
        message1: discord.Message = None,
        note: bool = True,
    ) -> Tuple[discord.Message, Optional[discord.Message], str]:
        if message1 is not None:
            if not (message1.author == self.bot.user and message1.embeds):
                raise ValueError("Malformed thread message.")

        elif message_id is not None:
            try:
                message1 = await self.channel.fetch_message(message_id)
            except discord.NotFound:
                raise ValueError("Thread message not found.")

            if not (message1.author == self.bot.user and message1.embeds):
                raise ValueError("Malformed thread message.")

        thread_logs = await self.bot.api.logs.find_one({"channel_id": str(self.channel.id)})
        if thread_logs is None:
            raise ValueError("Thread logs not found.")

        messages_logs = thread_logs["messages"]

        def get_data(msg_id):
            # internal function since this will be used twice
            linked_ids, type_, from_mod = next(
                (
                    (
                        data.get("linked_ids"),
                        data.get("type"),
                        data.get("author", {}).get("mod", False),
                    )
                    for data in messages_logs
                    if str(msg_id) in data.get("linked_ids", [])
                    or str(msg_id) == data.get("message_id", str())
                ),
                (None, str(), False),
            )
            return linked_ids, type_, from_mod

        if message1 is None:
            # still None, this is usually when using `delete` or `edit` command without providing ID
            # so in this case, we only search for thread message
            # Note or persistent note will not be returned if ID is not provided
            async for message1 in self.channel.history():
                linked_ids, type_, from_mod = get_data(message1.id)
                if from_mod and type_ not in ("system", "internal"):
                    break
            else:
                raise ValueError("Thread message not found.")
        else:
            linked_ids, type_, from_mod = get_data(message1.id)

        if linked_ids:
            async for msg in self.recipient.history():
                if str(msg.id) in linked_ids:
                    if not from_mod and not either_direction:
                        raise ValueError("Thread message not found.")
                    return message1, msg, type_
            raise ValueError("DM message not found.")

        if type_ == "system" and note:
            return message1, None, type_

        raise ValueError("Thread message not found.")
```

Jerrie-Aries:
Closing this since this is now outdated and there would be merge conflicts.

PR #3017, opened 2021-05-06
Create stat.py

NroxpTheDev:


StephenDaDev:
What exactly is the purpose of this?

NroxpTheDev:
Statcord To Get Their CPU Memory Guild And Server Count

onerandomusername (commented):
It looks like this pr would be better suited as a plugin. It doesn't add any additional features to the bot except for command tracking to a third party. 

I would also take a look at the [CONTRIBUTING.md](https://github.com/kyb3r/modmail/blob/master/.github/CONTRIBUTING.md) file for the requirements before contributing.

In the technical side of things,  it seems like this adds a dependency, but where are those edits?
I also don't see where this cog is loaded...?

I can't stress this enough, this should really be a plugin and not a built in feature of the bot.

lorenzo132:
Close this PR, this should not become a core function, request someone to make this a plugin or make it urself.

PR #3018, opened 2021-05-06
Update bot.py

NroxpTheDev:


onerandomusername:
Dupe of #3017, & should be a plugin.

PR #3021, opened 2021-05-06
Fix raw snippet markdown

onerandomusername:
If a snippet ended with a \`, it would mean that there would be a row of 4 \`\`\`\` in a row, causing the snippet raw text to have a small error in markdown formatting.

This fixes that, in addition to making the code block language `markdown`.


onerandomusername:
Hey, sorry. I noticed I pointed this towards the wrong branch, should I close this and open a new pr to the correct branch of `development`?

PR #3023, opened 2021-05-10
Don't remove reactions on pagination

onerandomusername:
This stops the bot from removing reactions when doing pagination, and now listens for both `reaction_add` and `reaction_remove` events, meaning it no longer has to remove reactions.

### DIFF
--- a/core/paginator.py
+++ b/core/paginator.py
@@ -134,18 +134,34 @@ class PaginatorSession:
             await self.show_page(self.current)
         while self.running:
             try:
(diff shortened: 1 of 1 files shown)


StephenDaDev:
Not removing the reactions after a set time will cause confusion on why it doesn't work.

onerandomusername:
> Not removing the reactions after a set time will cause confusion on why it doesn't work.

R.danny does this just fine. Its honestly how all bots should implement reaction pagination.

fourjr:
Non-issue. This will not be merged into modmail.

PR #3026, opened 2021-05-11
Update registry.json

NroxpTheDev:


fourjr (changes requested):
Invalid repository format. Refer to [Folder Structure](https://github.com/kyb3r/modmail/wiki/Plugins#folder-structure) here.

fourjr (changes requested):
Configuration keys should be editable through commands using a database partition. There should be no config keys hardcoded into plugin's code.

NroxpTheDev:
Ok

PR #3028, opened 2021-05-11
Fix breaking modmail internal thread cache

### LINKED ISSUE #3022: [BUG] It's possible to break modmail's internal thread cache and simultanously open many threads
**Describe the bug**
If `confirm_thread_creation` is used, a user can break the thread creation flow and somehow cause modmail's DM response to freeze. This then results in modmail sending as many `confirm_thread_creation` prompts as messages were sent during the freeze, and if the user reacts to each one, they can open one thread per message. 

**Bot Info**
Bot version (check with `@modmail about`): 3.9.4
Host method (Heroku, self-host, etc): self-hosted

**To Reproduce**
1. Who can reproduce (ex. anyone, owners)? Anyone
2. Where can it be reproduced (ex. in thread channels, recipient DM's)? DM's
3. Done what to cause the error?
 .1. DM modmail, wait for the `confirm_thread_creation` message to come up. 
 .2. Block modmail
 .3. React to create the thread and close the thread before modmail can respond
 .4. Unblock modmail, and while it's frozen (~10-15s depending), send a few messages. The effect is worse if you wait a short time between each one
 .5. Observe _many_ `confirm_thread_creation` messages that all work. 
4. Any recently made changes to your bot? Some print statements, and requests hitting a channel category instead of the guild's text channels. 

**Error Logs**
https://paste.pythondiscord.com/luyevadeyu.txt

**Screenshots**
https://cdn.discordapp.com/attachments/620607373828030464/840607305586049034/Screen_Shot_test1.png
https://cdn.discordapp.com/attachments/620607373828030464/840607475187449866/Screen_Shot_test2.png
https://cdn.discordapp.com/attachments/620607373828030464/840607476554792990/Screen_Shot_test3.png

**Additional context**

Jerrie-Aries:
__**About this PR:**__
File `thread.py`:
- Make sure there's no existing cancelled thread before returning `None` in method 'find'.
This possibly could prevent the bot from creating multiple threads when the user repetitively sends messages in DM on confirm thread creation.
- Use `await thread.close()` instead of `bot.loop.create_task()` to make sure the thread for the recipient is completely removed from cache (if any) before creating another one.
- Use `bot.loop.create_task()`  when sending "Cancelled" embed to recipient if the thread is cancelled on confirm thread creation.

Resolves #3022

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -1115,16 +1115,16 @@ class ThreadManager:
             try:
                 await thread.wait_until_ready()
             except asyncio.CancelledError:
-                logger.warning("Thread for %s cancelled, abort creating", recipient)
+                logger.warning("Thread for %s cancelled.", recipient)
                 return thread
             else:
-                if not thread.channel or not self.bot.get_channel(thread.channel.id):
+                if not thread.cancelled and (
+                    not thread.channel or not self.bot.get_channel(thread.channel.id)
+                ):
                     logger.warning(
                         "Found existing thread for %s but the channel is invalid.", recipient_id
                     )
-                    self.bot.loop.create_task(
-                        thread.close(closer=self.bot.user, silent=True, delete_channel=False)
-                    )
+                    await thread.close(closer=self.bot.user, silent=True, delete_channel=False)
                     thread = None
         else:
             channel = discord.utils.get(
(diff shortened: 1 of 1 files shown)


fourjr (approved):
LGTM

PR #3036, opened 2021-05-26
Update registry.json

NroxpTheDev:


fourjr (changes requested):
My comment from #3026 still stands. There is no point creating a new PR.
> Configuration keys should be editable through commands using a database partition. There should be no config keys hardcoded into plugin's code.

NroxpTheDev:
PR Has Been Closed

StephenDaDev:
Except it hasn't.

lorenzo132:
Ola @NroxpTheDev , this will be the last time we will mention this feature wont be added. Please do not make any new githubissues for this.

Issue #3041, opened 2021-06-01
[BUG] `delete` does not delete message for recipient


FloatingMilkshake:
**Describe the bug**
When a moderator deletes a message sent using `?delete`, it disappears from the thread channel but is not deleted from the recipient's DMs with the bot.

**Bot Info**
Bot version (check with `@modmail about`): 3.9.4
Host method (Heroku, self-host, etc): Heroku

**To Reproduce**
1. Who can reproduce (ex. anyone, owners)? Anyone
2. Where can it be reproduced (ex. in thread channels, recipient DM's)? Issue is present in recipient DMs, however to reproduce you must use a command in a thread channel
3. Done what to cause the error? Delete a mod-sent message, observe as it is deleted from the thread channel but not from the recipient's DMs
4. Any recently made changes to your bot? No changes, the bot is copied exactly from this repository
5. Errored

**Error Logs**
https://hastebin.cc/sihuwejobo

**Screenshots**
![zJ5JPYrDkS](https://user-images.githubusercontent.com/51096169/120257865-02d30e80-c25f-11eb-8bf7-43f087687d7a.png)
Note that this screenshot is from a different server than the attached logs and as such, user/server/bot names and IDs may not match. Issue is the same and is equally reproducible in both servers. Bot is up-to-date in both (version 3.9.4), hosted with Heroku in both.

Jerrie-Aries:
Please double check, make sure your bot has `View Audit Log` permission on your guild.

lorenzo132:
This same issue is for me on a 767k membered guild, has all needed perms including `View Audit Log`. So this should not be the problem @Jerrie-Aries

FloatingMilkshake:
I can confirm that my bot also has the `View Audit Log` permission.

StephenDaDev:
Unable to reproduce on the latest version, (the bot we use in the support server), VPS hosted.

StephenDaDev:
Any logs when it happens?

Jerrie-Aries:
> This same issue is for me on a 767k membered guild, has all needed perms including `View Audit Log`. So this should not be the problem @Jerrie-Aries
> I most likely think this has something to do with any form of caching fault

Sorry, my bad Renzo. I was just telling what the error log says, because it seemed like it came from here in `bot.py`:
```py
File "/app/bot.py", line 1455, in on_message_delete
    entry = await audit_logs.find(lambda a: a.target == self.user)
```
[...]

PR #3057, opened 2021-06-27
Hotfix: Deleting messages for thread recipients

lorenzo132:
reverted https://github.com/kyb3r/modmail/commit/beb73d8d82960429631c0aef5aea31f4f75f23b1
this causes messages to not delete in dm channels

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -697,10 +697,10 @@ class Thread:
         tasks = []
         if not isinstance(message, discord.Message):
             tasks += [message1.delete()]
-        elif message2 is not [None]:
+        if message2 is not [None]:
             for m2 in message2:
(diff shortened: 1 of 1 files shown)


Erisa:
Echoing my thoughts from #3041:
It's a great hotfix for now, however I have noticed one time where deleting a message said that it failed when it in fact succeeded.

A good longterm solution going forward would be what @Jerrie-Aries mentioned and appears to have committed to some dev fork. (https://github.com/Jerrie-Aries/modmail/commit/25aa1d4bf23bcb1a374c7f7e9a3eb3fa1cbb7573)

fourjr:
@Jerrie-Aries's solution will be used instead.

PR #3079, opened 2021-08-02
Add Python Discord's plugins to the registry

Akarys42:
This PR adds the following plugins:
- **Case insensitive snippets**: Allow snippets to be ran even if with the wrong case. For example, `?Dm-RePort` will be recognized as `?dm-report`.
- **Close message:** Add a `?closemessage` command that will close the thread after 15 minutes with a default message.
- **MDLink**: Generate a ready to paste link to the thread logs.
- **Reply cooldown**: Forbid you from sending the same message twice in ten seconds.
- **Tagging**: Add a `?tag` command capable of adding a `$message｜` header to the channel name.

The source code can be found in [our repository](https://git.pydis.com/modmail-plugins). I have the authorisation from all the authors to create this PR.

### DIFF
--- a/plugins/registry.json
+++ b/plugins/registry.json
@@ -1,4 +1,49 @@
 {
+    "case_insensitive_snippets": {
+        "repository": "python-discord/modmail-plugins",
+        "branch": "main",
+        "description": "Allow snippets to be ran even if with the wrong case. For example, ?Dm-RePort will be recognized as ?dm-report.",
+        "bot_version": "2.20.1",
+        "title": "Case insensitive snippets",
+        "icon_url": "https://i.imgur.com/kaJYlMB.png",
+        "thumbnail_url": "https://i.imgur.com/kaJYlMB.png"
+    },
+    "close_message": {
+        "repository": "python-discord/modmail-plugins",
+        "branch": "main",
+        "description": "Add a ?closemessage command that will close the thread after 15 minutes with a default message.",
+        "bot_version": "2.20.1",
+        "title": "Close message",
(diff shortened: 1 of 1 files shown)


RealCyGuy:
`?closemessage` can be achieved with `?alias add closemessage close 15m Feel free to open a new thread if you need anything else.`
`?mdlink` is just `?loglink` with some formatting.
Everything else seems fine though.

Akarys42:
Thank you for taking a look at this PR!

> ?closemessage can be achieved with ?alias add closemessage close 15m Feel
> free to open a new thread if you need anything else.
>
It accepts a duratiojnarguelnt that isn't possible to achieve using normal
aliases AFAIK

> ?mdlink is just ?loglink with some formatting.
>
I can remove it if you wish?

>

RealCyGuy:
I think those two plugins are too niche but the other plugins look more useful. If someone wants to install those plugins, they could always install it the long way.

lorenzo132:
i dont see any reason why those plugins cannot be added @RealCyGuy

fourjr:
Hey, thanks for the contribution.

We will be rejecting the following plugins:
- Case insensitive snippets: Feature is going to be added to v3.10.0 (currently on development branch) see #3080
- Tagging: Feature seems too niche

Akarys42:
Thank you! Will this be live right now or we shall wait for the next
release?

fourjr:
This should be live as soon as the registry cache expires. Might be live now, you can test it out on your instance with `?plugin registry mdlink`

PR #3080, opened 2021-08-03
Let snippets be invoked case-insensitively.

### LINKED ISSUE #3077: Let snippets be invoked case-insensitively.

It seems unintuitive that snippets cannot be invoked case-insensitively whilst commands can - all the plugin sub-commands even coerce the snippet name to be lower-case such that `?snippet FOO` is equivalent to `?snippet foo`.

`?snippet add FOO` followed by `?FOO` doesn't even work since a snippet by the name "foo" is added.

I'd like for `?FOO` to be equivalent to `?Foo` or `?foo` or any other combination of cases for that matter.

- [ ] Logviewer
- [ ] New config option

Making a plugin with a custom `process_commands`.

### Who will this benefit?

Everyone who uses snippets. I find myself accidentally capitalizing the first letter of the snippet accidentally only to have it not work.

### Additional Information

_No response_

Qwerty-133:
Closes #3077.

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -1132,8 +1132,11 @@ class ModmailBot(commands.Bot):
             cmd = message.content[len(self.prefix) :].strip()
 
             # Process snippets
-            if cmd in self.snippets:
-                snippet = self.snippets[cmd]
+            try:
+                snippet = self.snippets[cmd.lower()]
+            except KeyError:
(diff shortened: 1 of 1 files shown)


fourjr:
Merged into v3.10.0-dev4

PR #3081, opened 2021-08-08
Auto-trigger alias support

popeeyy:
Add support for alias commands in auto-triggers for easy editing of multiple triggers. Let me know if anything I did can be improved, I'm not very experienced with Python or any libraries such as discord.py.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project mostly adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html);
 however, insignificant breaking changes do not guarantee a major version bump, see the reasoning [here](https://github.com/kyb3r/modmail/issues/319). If you're a plugin developer, note the "BREAKING" section.
 
-# v3.9.5dev1
+# v3.9.5
 
-## Misc
+## Internal
 
 - Bumped discord.py to v1.7.3, updated all other packages to latest.
 - More debug log files are now kept.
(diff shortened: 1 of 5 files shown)


fourjr (changes requested):
This implementation is too basic and does not support multistep aliases (i.e. `reply hi && move appeals`), also arguments are not being passed for some reason (`reply hi` does not pass my `hi` argument). Is it possible to extract alias functionality to another function within Bot and call it (here and when invoking commands)?

Also, before committing, do remove your debug print statements and run `black .` for formatting checks. See out CONTRIBUTING.md for tips.

popeeyy:
> Is it possible to extract alias functionality to another function within Bot and call it (here and when invoking commands)?

I found the function that did this and I was looking for yesterday, but it is stuck on "if self._skip_check(message.author.id, self.user.id):" and otherwise works fine. What is skip_check as I don't see anything in the discord.py documentation and doing a search through all the code yields no results. Once I know what this is, I will fix that part and commit my changes and all the functionality should be brought back.

> Also, before committing, do remove your debug print statements and run `black .` for formatting checks. See out CONTRIBUTING.md for tips. PRs should also be pointed into development instead of master.

Thank you for letting me know, I will try to do that in the future. As you can see from the commits, I tried my best to look for all the debug statements and I thought I got them all out.

fourjr:
> Bot._skip_check
https://github.com/Rapptz/discord.py/blob/462ba84809f43349296c44cbe6468f631a00edab/discord/ext/commands/bot.py#L122

fourjr (approved):
lgtm thanks!

PR #3092, opened 2021-09-07
Fix `?contact` command bug where the channel is created inside random category.

### LINKED ISSUE #3091: [BUG]: ?contact xx#xxxx silently opens a thread outside of the Modmail category.
### Bot Version

v3.10.2

Heroku

### Error Logs

https://hastebin.cc/jotaguxaxi

### Screenshots

_No response_

### Additional Information

**Overview**: 

When `@contact xx#xxxx silently` is run, it opens a modmail thread outside of the Modmail category. I have reset the `main_category_id` and `fallback_category_id` to the Modmail category id, with no change.

**Background**:

Previously we used to be able to run the `?contact xx#xxxx silently` command and it would open a new channel without notifying the user. When this was run after the v3.10.2 release, it would open a channel without notifying the user, however in a public 'meta' category.

Running the `?contact xx#xxxx silent` command would open the thread in the correct location, however it would notify the user.

Request a fix, or a short-term workaround to prevent notification of the user until a `?reply` command is send in the respective channel.

Jerrie-Aries:
__**About this PR:**__
- Fix `?contact` command bug where the bot creates the thread channel inside a random category if `silent` or `silently` is passed in the command.

Resolve #3091

The cause of the issue is, there's a possibility an empty string is passed when converting the category using the `SimilarCategoryConverter` that makes the converter returns any category it found first.
I think the solution in this PR is dynamic enough to deal with that and user inputs.

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -1362,9 +1362,16 @@ class Modmail(commands.Cog):
         """
         silent = False
         if isinstance(category, str):
-            if "silent" in category or "silently" in category:
+            category = category.split()
+
+            # just check the last element in the list
+            if category[-1].lower() in ("silent", "silently"):
                 silent = True
-                category = category.strip("silently").strip("silent").strip()
+                # remove the last element as we no longer need it
+                category.pop()
+
+            category = " ".join(category)
+            if category:
                 try:
                     category = await SimilarCategoryConverter().convert(
                         ctx, category


fourjr (approved):
LGTM

dirtyformal:
@fourjr - Is there any chance we can get this merged into master soon, so we can fix a historical issue on our implementation?

Issue #3014, opened 2021-05-04
Anonymous Reply Highest role 


matrix2113:
**Describe the solution you'd like**
Update [this](https://github.com/kyb3r/modmail/blob/master/core/thread.py#L865) from `author.top_role` to be the highest hoisted role. 

**Who will this benefit**
Helps user so it doesn't show `^` or `*` to be the highest role in the embed. It'll show the next role under if it's hoisted like Admin or something.

StephenDaDev:
This would be beneficial to many users. A very large chunk of servers use roles like these or similar to give higher-level admin privileges, which often causes an issue. Another thing I have talked to people who want to do something where it just shows "Moderation Team" "Partnership Manager" etc instead of the actual rank of the staff member (but not the same for everyone).

onerandomusername:
Perhaps rather than editing that, adding a configuration for blacklisting roles from the highest role?

Or maybe a combination of both.

StephenDaDev:
It'd probably be easier for everyone involved to just use the highest hoisted.

fourjr:
Added in 3.10.0-dev5

PR #3093, opened 2021-09-07
Allow get_top_role to be optionally hoisted via config.

scragly:
PR #3014 introduced the behavior to only fetch hoisted top roles, which may be fine for most, but is an unexpected behavior to those who may have a secondary mod-mail server that doesn't bother to hoist the roles, and may specifically not wish to.

The result of no hoisted roles is predictably None being sent for anon reply titles and mod_tag reply footers.

image
image

In this PR, I've set this behavior to be configurable through a new config key: `use_hoisted_top_role`. Considering the majority of setups will be fine with it as-is, I've set it to True by default. 

I chose to use a simple `hoisted` flag on the util function rather than requiring `bot` be passed just to reference the config. This will also ensure the utility can be used in a generic manner in future for cases that may not want to rely on the same configuration value.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -30,7 +30,7 @@ __all__ = [
     "trigger_typing",
     "escape_code_block",
     "tryint",
-    "get_top_hoisted_role",
+    "get_top_role",
     "get_joint_id",
 ]
 
@@ -369,9 +369,11 @@ def tryint(x):
         return x
 
 
-def get_top_hoisted_role(member: discord.Member):
+def get_top_role(member: discord.Member, hoisted=True):
     roles = sorted(member.roles, key=lambda r: r.position, reverse=True)
     for role in roles:
+        if not hoisted:
+            return role
         if role.hoist:
             return role
 
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -918,7 +918,7 @@ class Modmail(commands.Cog):
 
             tag = self.bot.config["mod_tag"]
             if tag is None:
-                tag = str(get_top_hoisted_role(ctx.author))
+                tag = str(get_top_role(ctx.author, self.bot.config["use_hoisted_top_role"]))
             name = self.bot.config["anon_username"]
             if name is None:
                 name = tag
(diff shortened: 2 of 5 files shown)


fourjr:
Thanks for the PR. Haven't looked at it properly yet but could you add the new config option into config_help.json?

scragly:
Added, just let us know if the wording needs adjusting :+1: 
image

core/utils.py
```diff
@@ -369,9 +369,11 @@ def tryint(x):
         return x
 
 
-def get_top_hoisted_role(member: discord.Member):
+def get_top_role(member: discord.Member, hoisted=True):
     roles = sorted(member.roles, key=lambda r: r.position, reverse=True)
     for role in roles:
+        if not hoisted:
+            return role
```
Jerrie-Aries, line 376: Here you can just do:
```py
    if not hoisted:
        return member.top_role
```
before the `for` loop. Hmm, but the output still the same tho. 😅

And also, additionally, I'm suggesting if `hoisted` is `True` and no hoisted role is found after the `for` loop, it should fallback to `member.top_role`. Just to make sure this function return an appropriate role object.

scragly: > Here you can just do:

Thanks for the recommendation. It's a bit of a nitpick perhaps, as in discord.py, it's also iterating over the roles anyway in a similar manner. I'm inclined to leave it as is.

>  I'm suggesting if hoisted is True and no hoisted role is found after the for loop, it should fallback to member.top_role. Just to make sure this function returns an appropriate role object. 

Honestly, this should probably have been the default behavior back when the original PR change occurred, however I didn't want to assume there wasn't considerations behind why it wasn't. Instead, my intention is to not alter the existing behavior as was written by @fourjr originally, and just make it configurable. 

I don't want to creep the scope of this PR beyond this intention. If I did, I'd probably end up re-writing the entire project eventually just because I found things a whole bunch of things I think should be different. 

If the behavior you state is something you want, feel free to create a PR after this is merged.

fourjr (approved):
lgtm

PR #3098, opened 2021-09-13
Add a workflow to push to Docker hub on release

ChrisLovering:
This pushes an image to dockerhub with the tags latest, a short git sha, and the tag of the release.

The trigger for this workflow is when a release is marked as released.

This workflows requires two secrets to be made available to the repo:
- DOCKERHUB_USERNAME the username to login as
- DOCKERHUB_TOKEN a [personal access token](https://docs.docker.com/docker-hub/access-tokens/) with access to the modmail image

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -1,0 +1,46 @@
+name: Build and push to Docker hub
+
+on: 
+  release:
+    types: [released]
+
+jobs:
+  build-and-push:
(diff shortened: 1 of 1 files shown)


Akarys42 (approved):
For what my opinion is worth, that seems about right

viharm:
This merge request seems closed but the latest image on _Docker_ hub is still more than a year old. it appears to be based on _Modmail_ v3.9.5 (_discord.py_ v1.7.3).

PR #3103, opened 2021-09-18
Kill Command

Space-Turtle0:
This isn't much but it's just a simple way of killing the bot without the use of eval. (For those who don't know how to use Python yet) It's also a very simple way of ending the bot incase you need to do some maintenance.

### DIFF
--- a/cogs/utility.py
+++ b/cogs/utility.py
@@ -266,6 +266,7 @@ class Utility(commands.Cog):
     def cog_unload(self):
         self.bot.help_command = self._original_help_command
 
+    @commands.command()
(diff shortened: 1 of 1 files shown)


Atharv-Agarwal:
Adding a confirmation would be better.

fourjr:
Hey, for killing/restarting the bot, we prefer to recommend the use of outside tools such as within Heroku's disabling workers/restarting dynos. Using a sys.exit frequently on Heroku can result in the bot being permanently dead (until you manually restart dynos).

In addition, when the bot is killed "by accident" on discord, it might not be easy to login to a web dashboard/SSH session to restart the bot, this is made worse on patreon bots or bots that are hosted by others.

PR #3109, opened 2021-10-02
Fix fallback category error.

Jerrie-Aries:
- Fix issue where the bot doesn't create a new fallback category when the main category reaches the limit due to shadowing variable names.


fourjr (approved):
lgtm

PR #3111, opened 2021-10-09
Use new regex to properly match/parse channel topics.

Jerrie-Aries:
__About this PR:__
- New regex in `core/utils.py`.
- Add a `parse_channel_topic` function that uses the new regex to parse the channel topic and return a tuple of 'title', 'user ID', and 'other recipients IDs'.
- Remove couple of old regexes as they're no longer used.

With these new regex and methods, it would be easier to check and match the recipient ID.

Currently, in `core/thread.py`, line `1243`:
```py
channel = discord.utils.find(
                lambda x: str(recipient_id) in x.topic if x.topic else False,
                self.bot.modmail_guild.text_channels,
            )
```
..the bot just checks if the user ID is in the channel topic, regardless the string format. This PR solves that.

Please check.

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -1848,7 +1848,7 @@ class Modmail(commands.Cog):
                 and message.embeds[0].color.value == self.bot.main_color
                 and message.embeds[0].footer.text
             ):
-                user_id = match_user_id(message.embeds[0].footer.text)
+                user_id = match_user_id(message.embeds[0].footer.text, any_string=True)
                 other_recipients = match_other_recipients(ctx.channel.topic)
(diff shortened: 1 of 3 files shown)


Taaku18 (approved):
Everything seems to be in order to me (please resolve the comment above), but I'm not the most familiar with group chats functionalities so it's best for 4jr to review it as well.

Also, if possible, would you be able to implement a "Bot ID: xxxxx" section into the topics and have it checked when finding the channel by topic along with this PR since it's somewhat related?

Jerrie-Aries:
Thanks for the review. The suggested changes (regex) have been added.

fourjr (approved):
LGTM

PR #3128, opened 2021-12-05
Fix the docker image

ChrisLovering:
Since discord.py is installed via git in requirements.txt the Dockerfile needs to have git installed to download it.

Also, since lottie was added, to support stickers, this bumps the requirements, so they are available within the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM python:3.9-slim as py
 
 FROM py as build
 
-RUN apt update && apt install -y g++
+RUN apt update && apt install -y g++ git
 COPY requirements.txt /
 RUN pip install --prefix=/inst -U -r /requirements.txt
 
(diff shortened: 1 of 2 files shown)


Akarys42 (approved):
I could confirm those changes fixed the image locally

Agent-Fennec (approved):

Cordila (commented):
@ChrisLovering Could you give a quick explanation of all the extra reqs added and why they are needed? Taku needs that before he can merge

ChrisLovering:
I imagine this is quite of of date now.

The new requirements were to support the addition of stickers. I can take a look at rebasing onto development and updating the requirements list as needed.

Cordila:
I ran this with the suggested changes and it seemed to still work. @Taaku18 could you please review this when you are free?

Issue #3152, opened 2022-05-04
[BUG]: Modmail sets the channel topic separately from channel creation


sebkuip:
### Bot Version

v4.0 (all versions)

### How are you hosting Modmail?

Other

### Error Logs

N/A

### Additional Information

When a thread is created, through multiple steps it finally invokes a discord API call [here to make a channel](https://github.com/kyb3r/modmail/blob/77489be6a21cf4befeab1dc7fa2c2d2098104cbc/core/utils.py#L431). This only sets the name, category and permissions however.

The channel topic with the user ID is only set [here in thread.py](https://github.com/kyb3r/modmail/blob/77489be6a21cf4befeab1dc7fa2c2d2098104cbc/core/thread.py#L199)

This means it's a separate API call, which might pose issues with rate limits. One should specify the topic when creating the channel instead.

PR #3154, opened 2022-05-04
User ID topic is set on thread creation now. Fix for #3152

sebkuip:
It is an extra API call to separately set the channel topic instead of passing it when creating. This issue was described in #3152


fourjr:
Thanks for the contribution, merged together with #3166

fourjr (approved):

PR #3159, opened 2022-05-28
Update app.json

haydenbla123:


StarnightFox (changes requested):
you've put your actual bots token in here-

PR #3169, opened 2022-06-29
fix: Added missing ` char

ProfessorVarox:
- Added missing ` character in update_notifications.examples[0]


fourjr:
Duplicate of #3095

PR #3178, opened 2022-08-24
Add `unhandled_by_cog` parameter to `ModmailBot.on_command_error`

### LINKED ISSUE #3170: [BUG] `ModmailBot.on_command_error` does not respect cog's `cog_command_error` error handler.
### Bot Version

v4.0.0-dev16

Other

### Error Logs

None

### Screenshots

_No response_

### Additional Information

[PLUGIN] Both cog's `cog_command_error` and `ModmailBot.on_command_error` handlers catch the same exception.

**Reproduce:**
Create a plugin with a error handler.
```py
import discord

from discord.ext import commands

from core import checks
from core.models import PermissionLevel


class MyPlugin(commands.Cog):
    def __init__(self, bot):
        self.bot = bot

    async def cog_command_error(self, ctx, error):
        if isinstance(error, TypeError):
            embed = discord.Embed(color=self.bot.error_color, description=str(error))
            await ctx.send(embed=embed)
        return
    
    @commands.command(name="raise")
    @checks.has_permissions(PermissionLevel.OWNER)
    async def raise_error(self, ctx):
        raise TypeError("Error raised.")
```

**Expected result:**
The `ModmailBot.on_command_error` handler should not be called since the error has already been dealt with within the cog/plugin.

**Actual result:**
Both `cog_command_error` and `ModmailBot.on_command_error` handlers catch the exception.
And since `ModmailBot.on_command_error` doesn't handle `TypeError`, the error is raised and traceback is printed on console.

**Error:**
```py
2022-07-06 16:10:49 __main__[1547] - ERROR: Unexpected exception:
Traceback (most recent call last):
  File "/Users/Projects/modmail/.venv/lib/python3.9/site-packages/discord/ext/commands/core.py", line 200, in wrapped
    ret = await coro(*args, **kwargs)
  File "/Users/Projects/modmail/plugins/@local/moderation/moderation.py", line 309, in raise_error
    raise TypeError("Error raised.")
TypeError: Error raised.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/Projects/modmail/.venv/lib/python3.9/site-packages/discord/ext/commands/bot.py", line 1330, in invoke
    await ctx.command.invoke(ctx)
  File "/Users/Projects/modmail/.venv/lib/python3.9/site-packages/discord/ext/commands/core.py", line 995, in invoke
    await injected(*ctx.args, **ctx.kwargs)  # type: ignore
  File "/Users/Projects/modmail/.venv/lib/python3.9/site-packages/discord/ext/commands/core.py", line 209, in wrapped
    raise CommandInvokeError(exc) from exc
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: TypeError: Error raised.
```

Jerrie-Aries:
- Resolve #3170

As per title, this allows plugin developers to have their own custom exceptions and have more control on how to handle errors from their plugins without flooding the console, etc.

Example in plugin file:
- Specific command error handler.
```py
# imports

class MyPlugin(commands.Cog):
    def __init__(self, bot):
        self.bot = bot

    @commands.command()
    @checks.has_permissions(PermissionLevel.OWNER)
    async def test(self, ctx, *, value: str = None):
        if value == "1":
            raise TypeError("TypeError.")
        elif value == "2":
            raise commands.BadArgument("BadArgument.")
        else:
            await ctx.send("None.")

    @test.error
    async def test_error(self, ctx, error):
        if hasattr(error, "original"):
            # usually (I'm not really sure of this) if the exception is raised inside the command and the type of exception is not inherited
            # from .CommandError, the type of error here would be .CommandInvokeError and the original exception
            # can be retrieved from the .original attribute.
            error = error.original
        if isinstance(error, TypeError):
            await ctx.send(str(error))
        else:
            await self.bot.on_command_error(ctx, error, unhandled_by_cog=True)
```

- Cog error handler, this will catch every command error raised from the plugin.
```py
# imports

class MyPlugin(commands.Cog):
    def __init__(self, bot):
        self.bot = bot

    @commands.command()
    @checks.has_permissions(PermissionLevel.OWNER)
    async def test(self, ctx, *, value: str = None):
        if value == "1":
            raise TypeError("TypeError.")
        elif value == "2":
            raise commands.BadArgument("BadArgument.")
        else:
            await ctx.send("None.")

    async def cog_command_error(self, ctx, error):
        if hasattr(error, "original"):
            error = error.original
        if isinstance(error, TypeError):
            await ctx.send(str(error))
        else:
            await self.bot.on_command_error(ctx, error, unhandled_by_cog=True)
```

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -1511,7 +1511,17 @@ class ModmailBot(commands.Bot):
         logger.error("Ignoring exception in %s.", event_method)
         logger.error("Unexpected exception:", exc_info=sys.exc_info())
 
-    async def on_command_error(self, context, exception):
+    async def on_command_error(
+        self, context: commands.Context, exception: Exception, *, unhandled_by_cog: bool = False
+    ) -> None:
+        if not unhandled_by_cog:
+            command = context.command
+            if command and command.has_error_handler():
+                return
+            cog = context.cog
+            if cog and cog.has_error_handler():
+                return
+
         if isinstance(exception, (commands.BadArgument, commands.BadUnionArgument)):
             await context.typing()
             await context.send(embed=discord.Embed(color=self.error_color, description=str(exception)))


Cordila:
Does this mean the bot will handle all errors as it does right now as long as there is no error handler in the cog and if there is a error handler in the cog it will respect the cogs error handler over the eh of the bot?

Jerrie-Aries:
Yes Cordilla. The bot will still handle all errors like before.

fourjr:
LGTM Thanks

PR #3189, opened 2022-09-15
Update suggest plugin to modmail v4

RealCyGuy:
(first)


fourjr:
Thanks for updating. Will push this together with v4 release.

Jaffar93:
How to update

PR #3192, opened 2022-09-20
Update config.py

Boopup:


sebkuip:
Did…did you just change all the default values for the config, instead of just using the config command, and then push your fork as a PR?

honestly I’m impressed.

but seriously: use the config command instead of changing the default values in the files. Is a lot easier and less prone to break. Also won’t reset/break every update.

PR #3195, opened 2022-09-21
Add Alias Make/Create as Alias to Alias Add to resolve GH#3194

StephenDaDev:
"The ?snippet add command has ?snippet make/create as aliases, while the alias command does not, which will confuse users due to the lack of continuity between the commands."


Taaku18:
Thanks, merged for next release.

PR #3199, opened 2022-09-21
Dockerfile to use Python3.10 

Cordila:
This is to maintain consistency with runtime.txt

Refer to PR #3197


Taaku18:
Thanks, it was implemented and merged in #3228.

PR #3202, opened 2022-09-22
Edit permissions in bulk by using cog names GH#2900

### LINKED ISSUE #2900: Bulk Permissions override [Command request]
**Is your feature request related to a problem? Please describe.**
I want to bulk override permissions for command category
Now i need to do this manually for every commands and it is hard.

**Describe the solution you'd like**
Add command `?perms buld_override (command category/all commands) (perms level)`
It should override all commands in provided category

**Describe alternatives you've considered**
Maybe you can send me code how to do this using ?eval command.

**Who will this benefit**
Users who want to override permissions in more commands.

Cordila:
Closes #2900, a 2 year old feature request. 

Have tested this out properly and ironed out all bugs to my understanding. Attached a few screenshots to help understand better.

image
image
image
image

### DIFF
--- a/cogs/utility.py
+++ b/cogs/utility.py
@@ -37,6 +37,33 @@ from core.paginator import EmbedPaginatorSession, MessagePaginatorSession
 logger = getLogger(__name__)
 
 
+class ArgParser(commands.Converter):
+    async def convert(self, ctx: commands.Context, argument: str):
+        words = argument.split()
+        if len(words) < 2:
+            raise commands.BadArgument("Not enough arguments")
+
+        *name, user_or_role = words
+        second_argument = None
+
+        try:
+            second_argument = await utils.User().convert(ctx, user_or_role)
+        except commands.BadArgument:
+            pass
+
+        try:
+            second_argument = await commands.RoleConverter().convert(ctx, user_or_role)
+        except commands.BadArgument:
+            pass
+
+        if second_argument is None:
+            first_argument = argument
+        else:
+            first_argument = " ".join(name)
+
+        return first_argument, second_argument
+
+
 class ModmailHelpCommand(commands.HelpCommand):
     async def command_callback(self, ctx, *, command=None):
         """Overwrites original command_callback to ensure `help` without any arguments
@@ -972,7 +999,7 @@ class Utility(commands.Cog):
 
         For example:
         - `{prefix}alias add r reply`
-        - Now you can use `{prefix}r` as an replacement for `{prefix}reply`.
+        - Now you can use `{prefix}r` as a replacement for `{prefix}reply`.
 
         See also `{prefix}snippet`.
         """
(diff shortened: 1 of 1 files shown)


cogs/utility.py
```diff
@@ -1293,15 +1337,28 @@ async def permissions_override(self, ctx, command_name: str.lower, *, level_name
...
+
+        if type_ == "group":
+            for command in set(group.walk_commands()):
+                logger.info(
+                    "Updated command permission level for `%s` to `%s`.",
+                    command.qualified_name,
+                    level.name,
+                )
+                self.bot.config["override_command_level"][command.qualified_name] = level.name
+                await self.bot.config.update()
```
Jerrie-Aries, line 1350:
```suggestion
            update = False
            for command in set(group.walk_commands()):
                logger.info(
                    "Updated command permission level for `%s` to `%s`.",
                    command.qualified_name,
                    level.name,
                )
                self.bot.config["override_command_level"][command.qualified_name] = level.name
                update = True

            if update:
                await self.bot.config.update()
```

Calling `.config.update()` in the `for ...` loop repetitively could cause hitting the rate limit (at some point).
In my opinion, it would be a good practice if we only call it once after everything is set.

cogs/utility.py
```diff
@@ -1396,46 +1460,82 @@ async def permissions_remove(
...
+                    return await ctx.send(embed=embed)
+                for command in set(group.walk_commands()):
+                    name = command.qualified_name
+                    level = self.bot.config["override_command_level"].get(name)
+                    if level is None:
+                        continue
+                    else:
+                        logger.info("Restored command permission level for `%s`.", name)
+                        self.bot.config["override_command_level"].pop(name)
+                        await self.bot.config.update()
```
Jerrie-Aries, line 1525:
```suggestion
                update = False
                for command in set(group.walk_commands()):
                    name = command.qualified_name
                    level = self.bot.config["override_command_level"].get(name)
                    if level is None:
                        continue
                    else:
                        logger.info("Restored command permission level for `%s`.", name)
                        self.bot.config["override_command_level"].pop(name)
                        update = True

                if update:
                    await self.bot.config.update()
```

*Same as previous one.*

Jerrie-Aries (commented):
Took a quick peek. Looks good. 👍
Got a couple of suggestions there. Thoughts?

Not sure about the `.update_perms` tho 🤔.

*Update:*
One more thing, the group name is case sensitive and sometimes it could be multiple words too, if I remember correctly.

cogs/utility.py
Cordila: Whats the point of 
`update = False`
and then setting it to True every loop and then checking if its True and then updating? 

Wont it just work if the `await self.bot.config.update()` is removed from the loop?

Cordila:
> _Update:_ One more thing, the group name is case sensitive and sometimes it could be multiple words too, if I remember correctly.

Yeah group name is case sensitive, thats why I take it and put it in first letter capitalized, this should work for all inbuilt cogs. For plugins however, they will need to enter the exact class name if it isnt named as first letter capitalised. 

And no, class names cant be more than one word so I dont think the group name can be more than one word!

cogs/utility.py
Jerrie-Aries: `False` before the loop.
Set to `True` only if it enters the loop and after `config` is changed.
There probably will be the case where the group (cog) is empty (without commands) tho, so the `update` will remain `False` and `.bot.config.update()` will not be executed.

Cordila: Ah makes sense!

Cordila:
> Yes, class name cannot have multiple words. But cog name could have multiple words (by specifying the `name` kwarg when creating the class). `bot.get_cog` will look for cog name instead of class name.

Cant they just use "cog name" 😕
I feel like to allow more than 1 word cog names will be very difficult.

PR #3233, opened 2022-12-04
Fix typo #3210

### LINKED ISSUE #3210: [BUG]: Scheduled close missing word "in".
### Bot Version

4.0.1

PM2

### Error Logs

boop

### Screenshots

_No response_

### modmail.py
line 449

        embed = discord.Embed(
            title="Scheduled close",
            description=f"This thread will close {silent}{human_delta}.",
            color=self.bot.error_color,
        )

Cordila:
Closes #3210 by fixing the typo.


cogs/modmail.py
```diff
@@ -448,7 +448,7 @@ async def send_scheduled_close_message(self, ctx, after, silent=False):
 
         embed = discord.Embed(
             title="Scheduled close",
-            description=f"This thread will close {silent}{human_delta}.",
+            description=f"This thread will close in {human_delta} {silent}.",
```
Taaku18, line 451: There are some spacing issues here: always an extra space before the period and an extra space before "silently" when silently close.

I suggest changing the string to: `f"This thread will{' silently' if silent else ''} close in {human_delta}."` then remove [L447](https://github.com/modmail-dev/modmail/pull/3233/files#diff-bdc24e11b743f28e1f3139a9a78b2ee2c7bd0c71b132e396d57841e9998ccd90R447).

Taaku18 (changes requested):

cogs/modmail.py
StephenDaDev: I agree this format is better than both the current and proposed, makes it more polished for end user

Taaku18 (dismissed):

PR #3239, opened 2023-01-07
Fix replies in dms not being sent

### LINKED ISSUE #3238: Allow "user replying" messeges to be sent to the thread

In a modmail thread, if the user sends a message but replies to a bot message, it won't be sent to the thread, confusing the users.

How the user sees the threads
How the staff sees the thread

I think it would be great if the bot sent to the thread all the messages the user sends, even if it's a normal message or a replying message. 

- [ ] Logviewer
- [ ] New config option

If the solution above can't be done, it would be great to indicate with a ❌ emoji that the message wasn't sent.

### Who will this benefit?

Modmail users mostly, as it can confuse them.

### Additional Information

_No response_

RealCyGuy:
In discord.py v2, a new message type [`discord.MessageType.reply`](https://discordpy.readthedocs.io/en/stable/api.html?highlight=messagetype#discord.MessageType.reply) was added so we have to check for that as well.

Fixes #3238

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -888,7 +888,7 @@ class ModmailBot(commands.Bot):
             return
         sent_emoji, blocked_emoji = await self.retrieve_emoji()
 
-        if message.type not in {discord.MessageType.default, discord.MessageType.reply}:
+        if message.type not in [discord.MessageType.default, discord.MessageType.reply]:
             return
 
         thread = await self.threads.find(recipient=message.author)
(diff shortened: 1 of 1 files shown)


StephenDaDev:
@Taaku18 If you could look at this as well its a pretty big issue

Taaku18 (approved):
Looks good, will merge now.

PR #3242, opened 2023-01-20
Optimized "blocked" command

khakers:
This fixes the issues those with large blocklists were having when attempting to use the  `blocked` command. Tested with 15k blocked user and had no performance problems (besides the memory usage going up by over 20MB).

There are some more optimizations I can make to this method that'll either end up being adding to this PR or done in the future separately depending on my time constraints.


lorenzo132:
Can confirm this solves the issue

StephenDaDev:
Also confirmed

PR #3243, opened 2023-01-21
Lints: update to setup Python v4

SpencerIsGiddy:
Updates setup python to v4 in lints.yml


raidensakura:
I don't think this PR is necessary, Modmail does not support 3.11 yet so there's no need to include 3.11 in action tests, it's only going to extend runtime.

Cordila:
> I don't think inclusion of 3.11 is necessary, Modmail does not support it yet so there's no need to include it in action tests, it's only going to extend runtime.

I had to change it to 3.11 locally to allow certain plugins of mine to work. There is no harm imo of staying up to date

GiddyGoatGaming:
> I don't think inclusion of 3.11 is necessary, Modmail does not support it yet so there's no need to include it in action tests, it's only going to extend runtime.

Ok. I reverted the change

Taaku18 (approved):

PR #3244, opened 2023-01-25
Add dm_channel_id field to log entry

khakers:
Adds a `dm_channel_id` field to the mongodb log entry document containing the id of the dm channel between the bot and the recipient. 

DM channel ids are random and we can't determine them with the information currently in the database. By adding this field, we'll enable features like giving users the link to a specific message, if, for instance, they wanted to report it to discord.

### DIFF
--- a/core/clients.py
+++ b/core/clients.py
@@ -568,6 +568,8 @@ class MongoDBClient(ApiClient):
     async def create_log_entry(self, recipient: Member, channel: TextChannel, creator: Member) -> str:
         key = secrets.token_hex(6)
 
+        dm_channel = await recipient.create_dm()
+
         await self.logs.insert_one(
             {
                 "_id": key,
@@ -578,6 +580,7 @@ class MongoDBClient(ApiClient):
                 "channel_id": str(channel.id),
                 "guild_id": str(self.bot.guild_id),
                 "bot_id": str(self.bot.user.id),
+                "dm_channel_id": str(dm_channel.id),
                 "recipient": {
                     "id": str(recipient.id),
                     "name": recipient.name,


sebkuip:
The message link uses the user ID. I don't think there's any way to obtain a channel ID from a DM, assuming that there even is one.

khakers:
> The message link uses the user ID. I don't think there's any way to obtain a channel ID from a DM, assuming that there even is one.
Links to direct messages don't use the user ID, they have their own channel IDs. That's the whole reason I made this PR. So if you want to create a link to a message you can take `dm_channel_id`/`message_id` and have a working link.

RealCyGuy:
DM message links don't use the dm channel id, they just use user ids (afaik).

```
https://discord.com/channels/@me/{user id}/{message id}
https://discord.com/channels/@me/954169091734765638/1070603535705714739
```

khakers:
image
For reference, `1` is my user id, `2` is the user id of the bot. If someone could explain to me how to derive the channel id from either of these two values, I would greatly appreciate it.

StephenDaDev:
This (DM Channel ID) used to be shown upon thread open but was removed on bump to v4. If you look back far enough its probably possible to see how it was done
Screen Shot 2023-03-13 at 7 17 35 PM


I'll take a look and see what I can find on the matter

StephenDaDev:
https://github.com/modmail-dev/modmail/pull/2966/commits/d24501bc2550de890b5389b8debc000bccad13b1

core/clients.py
```diff
@@ -578,6 +578,7 @@ async def create_log_entry(self, recipient: Member, channel: TextChannel, creato
                 "channel_id": str(channel.id),
                 "guild_id": str(self.bot.guild_id),
                 "bot_id": str(self.bot.user.id),
+                "dm_channel_id": str(recipient.dm_channel.id),
```
Taaku18, line 581: I believe it's possible for `dm_channel` to be None--more likely with larger servers. As this function gets called before sending the first message to the user, this could potentially be an issue.

You probably want to use `dm_channel = await recipient.create_dm()` beforehand. It returns `dm_channel` if it's cached, or otherwise it will fetch from Discord.

Taaku18 (changes requested):

PR #3247, opened 2023-03-02
Added the option for secure_plugins_only

sebkuip:
This .env option disables downloading plugins that are not in the registry. This way, plugins can be kept secure/verified on specific platforms.


cogs/plugins.py
```diff
@@ -302,6 +302,14 @@ async def parse_user_input(self, ctx, plugin_name, check_version=False):
             plugin = Plugin(user, repo, plugin_name, branch)
 
         else:
+            if not self.bot.config.get("registry_plugins_only", False):
```
Jerrie-Aries, line 305: Sorry for late comment.

Isn't this bit here supposed to be
```suggestion
            if self.bot.config.get("registry_plugins_only", False):
```
???

Because currently, I cannot add local plugins without setting anything for this config in `.env`.

Taaku18: Yes, you're right, my mistake for hastily merging this PR. I have pushed a fix to master.

PR #3252, opened 2023-03-12
Import/Export backups of collections in the database

### LINKED ISSUE #3248: [FR]: Ability to import and export settings

I have set up Modmail multiple times both with self-hosting methods and through the Patreon method and talking with @Taaku18 this feature hasn't actually been requested as of yet.

Having the ability to;

- [ ] Import settings of a currently running Modmail instance with an external file formatted as JSON sent via Discord.
- [ ] Import settings of either a currently running or not running Modmail instance directly from a Git-based provider (similar to how you load plugins)
- [ ] Export settings of a currently running Modmail instance to a JSON file downloadable via Discord.
- [ ] Export settings of a currently running Modmail instance to a Git-based repository (Providing you have given your instance a token that allows Modmail to interface with a Git-based service)
- [ ] Having error checking for settings that rely on snowflakes and data that is specific to an ID.

- [ ] Logviewer
- [ ] New config option

Due to what I'm asking for, the only alternative solution for the moment is to manually change each setting, providing you remember all of them and due to how configurable Modmail is, there is more chance of a user letting a setting fall through the cracks.

### Who will this benefit?

Honestly? I'm unsure as to who this would benefit outside of myself but I'm sure if the feature got implemented, there would be a small user base for such a feature who are constantly in need to set up multiple Modmail instances, this would just ease the process for people who have their Modmail's set up in a specific way.

### Additional Information

_No response_

Cordila:
Closes #3248 

I dont feel the need to have import exports to a git based provider so this PR only works with JSON files sent via discord.

[p]export <collection name>

[p]import <attach the file>

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -2201,7 +2201,7 @@ class Modmail(commands.Cog):
 
     @commands.command(name="import")
     @checks.has_permissions(PermissionLevel.ADMINISTRATOR)
-    async def import_backup(self,ctx):
+    async def import_backup(self, ctx):
         """
         Import a backup from a json file.
 
@@ -2215,9 +2215,15 @@ class Modmail(commands.Cog):
             await attachment.save(attachment.filename)
             file = discord.File(attachment.filename)
             collection_name = os.path.splitext(attachment.filename)[0]
-            await ctx.send(f"This will overwrite all data in the {collection_name} collection. Are you sure you want to continue? (yes/no)")
+            await ctx.send(
+                f"This will overwrite all data in the {collection_name} collection. Are you sure you want to continue? (yes/no)"
+            )
             try:
-                msg = await self.bot.wait_for("message",timeout=30,check=lambda m: m.author == ctx.author and m.channel.id == ctx.channel.id)
+                msg = await self.bot.wait_for(
+                    "message",
+                    timeout=30,
+                    check=lambda m: m.author == ctx.author and m.channel.id == ctx.channel.id,
+                )
                 if msg.content.lower() == "yes":
                     success_message = await self.bot.api.import_backups(collection_name, file)
                     await ctx.send(success_message)
@@ -2230,8 +2236,6 @@ class Modmail(commands.Cog):
 
         else:
             return await ctx.send("Please attach 1 json file.")
-
-
 
 
 async def setup(bot):
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -1,5 +1,6 @@
 import asyncio
 import re
+import os
 from datetime import datetime, timezone
 from itertools import zip_longest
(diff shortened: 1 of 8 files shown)


cogs/modmail.py
```diff
@@ -2186,6 +2187,56 @@ async def isenable(self, ctx):
 
         return await ctx.send(embed=embed)
 
+    @commands.command(name="export")
+    @checks.has_permissions(PermissionLevel.ADMINISTRATOR)
+    async def export_backup(self, ctx, collection_name):
+        """
+        Export a backup of a collection in the form of a json file.
+
+        {prefix}export <collection_name>
```
Taaku18, line 2196: The two usage examples, on L2196 and L2210, are unnecessary, since discord.py auto generates the usage spec. However, it might be useful to note in the import command's help doc that you should attach the backup file to the message.

cogs/modmail.py
```diff
@@ -2186,6 +2187,56 @@ async def isenable(self, ctx):
 
         return await ctx.send(embed=embed)
 
+    @commands.command(name="export")
+    @checks.has_permissions(PermissionLevel.ADMINISTRATOR)
+    async def export_backup(self, ctx, collection_name):
+        """
+        Export a backup of a collection in the form of a json file.
```
Taaku18, line 2194: As noted in my other comment, we should store the backup as BSON file instead of JSON. Please change the references to BSON files instead.

cogs/modmail.py
```diff
@@ -2186,6 +2187,56 @@ async def isenable(self, ctx):
 
         return await ctx.send(embed=embed)
 
+    @commands.command(name="export")
+    @checks.has_permissions(PermissionLevel.ADMINISTRATOR)
+    async def export_backup(self, ctx, collection_name):
```
Taaku18, line 2192: Instead of exporting a collection, most users doesn't know what even are collections. Could you change it so that it exports everything (into one file) and imports them as well?

core/clients.py
```diff
@@ -751,6 +759,29 @@ async def get_user_info(self) -> Optional[dict]:
                 }
             }
 
+    async def export_backups(self, collection_name: str):
+        coll = self.db[collection_name]
+        documents = []
+        async for document in coll.find():
+            documents.append(document)
+        with open(f"{collection_name}.json", "w") as f:
```
Taaku18, line 767: Don't write to disk, that may lead to potential issues. Use a `BytesIO` instead.

Taaku18 (changes requested):

cogs/modmail.py
Cordila: You mean export config/logs and any plugins into 1 bson file? The issue with that would be the bot wont be able to send the large files.

PR #3254, opened 2023-03-12
Reminder Plugin

martinbndr:
Created a reminder plugin

### DIFF
--- a/plugins/registry.json
+++ b/plugins/registry.json
@@ -35,11 +35,11 @@
         "icon_url": "https://i.imgur.com/qtE7AH8.png",
         "thumbnail_url": "https://i.imgur.com/qtE7AH8.png"
     },
-	"reminder": {
+    "reminder": {
         "repository": "martinbndr/kyb3r-modmail-plugins",
         "branch": "master",
         "description": "Let´s you create reminders.",
-        "bot_version": "4.0.0",          
+        "bot_version": "4.0.0",
         "title": "Reminder",
         "icon_url": "https://raw.githubusercontent.com/martinbndr/kyb3r-modmail-plugins/master/reminder/logo.png",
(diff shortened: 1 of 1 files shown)


plugins/registry.json
```diff
@@ -34,6 +34,15 @@
         "title": "Suggest stuff.",
         "icon_url": "https://i.imgur.com/qtE7AH8.png",
         "thumbnail_url": "https://i.imgur.com/qtE7AH8.png"
+    },
+	"reminder": {
```
sebkuip, line 38: Indentation error here. The name of the plugin should be indented only once (not twice)

martinbndr: image
Looks fine for me on my end hmm

martinbndr: Also tried a different program, its showing like this.
Do you have an idea wihy its wrong on github?
image

StephenDaDev:
The plugin does not seem to install correctly onto a stock modmail bot. See logs: [https://hastebin.cc/yalopivare](https://hastebin.cc/yalopivare). While I could probably get it to work without erroring on me, plugins designed to be in the registry should work OOB, as most people using modmail don't really understand the technical side of things.

plugins/registry.json
StephenDaDev: Yeah it looks fine in VSC, (incl. the GH hosted VSC) not sure why it appears weird on the normal instance

martinbndr:
image
Hmm strange, I tested installation without any plugins and it works

martinbndr:
Update: now it should work, @StephenDaDev  Sry I made a stupid mistake

lorenzo132:
Can confirm it now works

StephenDaDev (approved):
Tested as working, formatting is all good

Taaku18 (approved):

PR #3255, opened 2023-03-12
Update links to the new org/repo.

japandotorg:


japandotorg:
ah looks like someone already made one rip

Issue #3115, opened 2021-10-31
[BUG]: Incorrect date type stored in MongoDB


gdude2002:
### Bot Version

v3.10.2

### How are you hosting Modmail?

Patreon

### Error Logs

N/A

#### Sample thread document

![Sample thread document](https://user-images.githubusercontent.com/204153/139593320-21efa98c-78ee-428f-8770-a37f5f8ba7ae.png)

#### Metabase error

![Metabase error](https://user-images.githubusercontent.com/204153/139593362-81e0fe34-5b53-494d-83ee-f758c2711fc7.png)

### Additional Information

This isn't so much a user-level bug as it is a data integrity issue.

When attempting to make use of the data stored in the MongoDB database managed by the modmail bot, it's impossible to filter or group based on dates because fields such as `created_at` are strings, rather than actual valid MongoDB dates.

A fix (and database migration step) for this issue is probably a good idea - not just so that one can perform statistical analysis against the database using external tools, but also just to improve overall data integrity.

PR #3257, opened 2023-03-13
Add config to manage how long logs are stored

### LINKED ISSUE #2942: Can you please add logs lifetime config option?
Hello, can you please add logs_lifetime config option? That you can set time (in ms) that how long logs will be stored in database. 
After that time remove logs from database.

By default this option should be disabled (seted to 0 = dont delete logs)

Atharv-Agarwal:
Closes #2942 

This pr adds a config option called `log_expiration` which can be configured in the same way as `thread_cooldown` etc. 
According to the `closed_at` field this config option will delete logs which have passed the time specified in `log_expiration`. 

This will work retrospectively.

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -1736,8 +1736,7 @@ class ModmailBot(commands.Bot):
                         next_expiry = closed_at + log_expire_after
                         return await discord.utils.sleep_until(next_expiry)
 
-        for log in expired_logs:
-            await self.db.logs.delete_one({"_id": log["_id"]})
+        await self.db.logs.delete_many({"closed_at": {"$lte": str(expiration_datetime)}})
         logger.info(f"Deleted {len(expired_logs)} expired logs.")
 
         fetch = await self.db.logs.find().sort("closed_at", 1).limit(1).to_list(None)
--- a/bot.py
+++ b/bot.py
@@ -12,7 +12,7 @@ import struct
 import sys
 import platform
 import typing
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 from subprocess import PIPE
 from types import SimpleNamespace
 
@@ -633,6 +633,7 @@ class ModmailBot(commands.Bot):
 
         self.post_metadata.start()
         self.autoupdate.start()
+        self.log_expiry.start()
         self._started = True
 
     async def convert_emoji(self, name: str) -> str:
@@ -657,7 +658,6 @@ class ModmailBot(commands.Bot):
         return self.get_user(id) or await self.fetch_user(id)
 
     async def retrieve_emoji(self) -> typing.Tuple[str, str]:
-
         sent_emoji = self.config["sent_emoji"]
         blocked_emoji = self.config["blocked_emoji"]
 
@@ -731,7 +731,6 @@ class ModmailBot(commands.Bot):
         if isinstance(author, discord.Member):
             for r in author.roles:
                 if str(r.id) in self.blocked_roles:
-
                     blocked_reason = self.blocked_roles.get(str(r.id)) or ""
 
                     try:
@@ -790,7 +789,6 @@ class ModmailBot(commands.Bot):
         channel: discord.TextChannel = None,
         send_message: bool = False,
     ) -> bool:
-
         member = self.guild.get_member(author.id)
         if member is None:
             # try to find in other guilds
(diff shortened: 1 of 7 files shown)


Cordila:
I have tested this extensively but it would be better if someone could review this to make sure its efficient. Also since setting the config option will immediately start deleting old logs not sure if we should have some kind of safeguard?

StephenDaDev:
I think a safeguard is probably a good measure just to make sure nobody does anything they'd rather not do

Cordila:
> I think a safeguard is probably a good measure just to make sure nobody does anything they'd rather not do

Actually if you look at it, the config option is set after the bot is running, they will need to restart the bot for it to come into effect. In this time they can remove the config option to prevent logs from being deleted. However vise versa if the logs are being deleted and they remove the config option, the deletion will stop without needing a restart. 

Maybe this is enough?

StephenDaDev:
@lorenzo132 What do you think?

lorenzo132:
> @lorenzo132 What do you think?

I'm a fan of the logs expiration config, specialy on a company wise level this is a very nice feature to have.

Cordila:
Used `delete_many` instead of looping `delete_one` based on recommendation by @khakers

bot.py
```diff
@@ -1714,6 +1712,41 @@ async def before_autoupdate(self):
             self.autoupdate.cancel()
             return
 
+    @tasks.loop()
+    async def log_expiry(self):
```
khakers, line 1716: Not a showstopper, but this method seems incredibly overcomplicated for what it *needs* to do. I would say it could be trimmed to down to deleting the logs at line 1723 (you get the amount deleted from this so you can log it) and then repeating again after a fixed time, which is how these tasks are typically done anyways. Doing this would make the code much more readable and efficient (though I don't think the extra database calls this currently has would ever actually matter at the scale it's likely to be used at).

Cordila: I don't think it would make it more efficient because instead of running every x minutes and not having any logs to delete, the task will only run when a log needs to be deleted.

khakers: It's probably going to be mostly irrelevant in efficiency either way (running one query on the database once an hour or day or something that only returns deletion data is basically meaningless, as are basically all the other ones here). I was mostly referring to code readability that you would get by removing all the exit points in this method (and basically removing most of the code) and not duplicating the same query multiple times.

For reference, this is what I'm talking about (untested code):

```python
async def expire_logs(self, log_expire_after):
        now = datetime.utcnow()
        expiration_datetime = now - log_expire_after
        expired_logs = await self.db.logs.delete_many({"closed_at": {"$lte": str(expiration_datetime)}})

        logger.info(f"Deleted {len(expired_logs.deleted_count)} expired logs.")
 ```

imo this is much easier to read and the only functionality you lose is precise scheduling (and you gain being able to easily run it with a command if you wanted)

Comment (Taaku18) [reply bot.py:1716]: I agree that the majority of the function looks overengineered. I think you can use a constant period loop over the function, and then directly delete from the database based on a query would work better and be more efficient.

We can have the task auto-repeat every hour:
`@tasks.loop(hours=1, reconnect=False)`

File: bot.py [hunk 2]
```diff
@@ -1714,6 +1712,41 @@ async def before_autoupdate(self):
...
+        log_expire_after = self.config.get("log_expiration")
+        if log_expire_after == isodate.Duration():
+            return self.log_expiry.stop()
+
+        now = datetime.utcnow()
```
Taaku18, line 1721:
```suggestion
        now = discord.utils.utcnow()
```

For consistency with the rest of the codebase, we use discord.py's `utcnow()`, which includes the timezone info.

bot.py
```diff
@@ -1714,6 +1712,41 @@ async def before_autoupdate(self):
...
+        if not expired_logs:
+            fetch = await self.db.logs.find().sort("closed_at", 1).limit(1).to_list(None)
+            if fetch:
+                for x in fetch:
+                    if x["closed_at"] > str(expiration_datetime):
+                        closed_at = datetime.strptime(x["closed_at"], "%Y-%m-%d %H:%M:%S.%f%z")
+                        next_expiry = closed_at + log_expire_after
+                        return await discord.utils.sleep_until(next_expiry)
+
+        await self.db.logs.delete_many({"closed_at": {"$lte": str(expiration_datetime)}})
```
Taaku18, line 1739: `closed_at` is stored as a string, see #3115. You cannot directly compare `closed_at` using `lte` string comparison. You'll need to use an MongoDB aggregation pipeline function (most likely `$dateFromString`) then compare as date types.

Taaku18 (changes requested):

bot.py
Cordila: The string compare does work as its comparing both of them as strings. The issue there is a database integrity issue where we are storing date time as a string, this would require a overhaul of the existing databases

Cordila:
All requested changes have been done and tested.

core/config_help.json
```diff
@@ -373,6 +373,17 @@
       "To disable thread cooldown, do `{prefix}config del thread_cooldown`."
     ]
   },
+  "log_expiration": {
+    "default": "Never",
+    "description": "The duration closed threads will be stored within the database before deletion. Logs that have been closed for longer than this duration will be deleted automatically.",
+    "examples": [
+      "`{prefix}config set thread_cooldown P12DT3H` (stands for 12 days and 3 hours in [ISO-8601 Duration Format](https://en.wikipedia.org/wiki/ISO_8601#Durations))",
+      "`{prefix}config set thread_cooldown 3 days and 5 hours` (accepted readable time)"
```
Jerrie-Aries, line 381:
```suggestion
      "`{prefix}config set log_expiration P12DT3H` (stands for 12 days and 3 hours in [ISO-8601 Duration Format](https://en.wikipedia.org/wiki/ISO_8601#Durations))",
      "`{prefix}config set log_expiration 3 days and 5 hours` (accepted readable time)"
```

bot.py
```diff
@@ -1697,6 +1695,18 @@ async def before_autoupdate(self):
             self.autoupdate.cancel()
             return
 
+    @tasks.loop(hours=1, reconnect=False)
+    async def log_expiry(self):
+        log_expire_after = self.config.get("log_expiration")
+        if log_expire_after == isodate.Duration():
+            return self.log_expiry.stop()
```
Jerrie-Aries, line 1702: What about just doing `return` here without stopping the loop?

That way if we set the `log_expiration` later, the loop will just continue to execute the next block without us needing to restart the bot.

Jerrie-Aries (commented):
There were copy and paste errors in `config_help.json`, I believe.

Also a suggestion about not stopping the loop if the config for it is not set.

bot.py
Cordila: The reason why I stopped the loop instead of just returning is that the restart serves as a failsafe mechanism if someone accidentally sets a log expiration, and it just so happens to loop right before they can change it. An edge case scenario really.

Jerrie-Aries:
I did a couple of quick tests. Seems working fine for me.

Taaku18 (approved):
Thank you for making the requested changes! Looks good.

PR #3262, opened 2023-03-21
group for imports

WebKide:
made 3 groups of imports for ease of maintenance


Taaku18:
We use black to format our code, and use this import order: 1. Built in imports 2. Dependency imports 3) Discord imports (project preference) 4. Local imports. Raw imports on top, from... import... after, sorted alphabetically.

There's no need to change this.

PR #3266, opened 2023-03-21
Arranging imports in 3 groups

WebKide:
well, this will help to maintain the code, add or remove modules and such

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -1,21 +1,22 @@
-import logging
-import re
-import sys
-import os
-from difflib import get_close_matches
+# Plain imports
+import logging, os, re, sys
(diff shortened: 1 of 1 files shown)


raidensakura (commented):
First of all, you should merge all your PRs into one if they're fixing the same issue. Multiple separated PRs like the ones you made only make it harder to keep track and handle.

Secondly, like Taku said in discord dev chat, Python has import sorting convention that your PRs do not follow.

I suggest looking into isort, it's a tool to automatically sort Python imports, and suggest on how it can be used in our current project workflow in your PR for better consistency (ie: pre-commit hooks, workflow steps etc)

PR #3272, opened 2023-03-28
Improve tagging of docker images

Huskydog9988:
This creates a number of tags for the docker images, allowing for users to properly follow semver.

To breakdown what tags are created:'
`nightly`, ex `nightly` (if you run on schedule)
`{{version}}`, ex `4.0.0`
`{{major}}.{{minor}}`, ex `4.0`
`branch`, ex `master` (when you do on push)
Also does a tag for the git sha

See companion pr: https://github.com/modmail-dev/logviewer/pull/71

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,9 +1,8 @@
-
 name: Create and publish a Docker image
 
 on:
   push:
-    branches: ['master']
+    branches: ["master"]
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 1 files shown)


Taaku18 (changes requested):
Hello, thanks for the PR. I'm not too familiar with the GitHub action for Docker. But I think there are a few issues with your commit:

The action is only set to run for "push to master", while all the `type=semver` requires "on tag push" events, and `type=schedule` requires "schedule" events. 

`type=ref,event=pr` looks to be unnecessary, since we don't want an image for every PR, and there's no event for PR.

And I don't think `type=raw,value=latest,enable=${{ github.event.release.prerelease == false }}` makes sense, since this action isn't part of a release event. `type=raw,value=latest,enable={{is_default_branch}}` would probably work better.

PR #3273, opened 2023-04-03
confirm_thread_creation Buttons instead of reactions

martinbndr:
This pr replaces reactions with buttons for the confirm_thread_creation feature.
Buttons should be more userfriendly and stable for this feature. Changes tested, should work 100%


Taaku18 (approved):
Looks good, thanks for the PR. I updated some relevant messages regarding the change, and removed the changelog entry–as it'll be added later once merged.

PR #3276, opened 2023-04-10
Add `rename` to `registry.json`

Nicklaus-s:


StephenDaDev:
Also not sure what the icon is

Nicklaus-s:
I've added a permission group check, see line [16](https://github.com/Nicklaus-s/modmail-plugins/blob/aad380de2ebabbae61026b280a9879bac7d550dd/rename/rename.py#L16) & made sure to include a fallback exception, see line [40](https://github.com/Nicklaus-s/modmail-plugins/blob/aad380de2ebabbae61026b280a9879bac7d550dd/rename/rename.py#L40-#L42). I'm not exactly sure if this is what you were asking for, but I believe I've fixed everything.

I've also changed the attached icon.

Nicklaus-s:
chat is it true this from april

StephenDaDev (approved):

StephenDaDev:
@Taaku18 Recommending for addition in next update

StephenDaDev:
_clicked the wrong button_

Taaku18 (approved):
Thanks for the addition, will be added to the registry in the next update.

PR #3278, opened 2023-04-11
[Fix] disable new command not updating db config

martinbndr:
Steps to reproduce:
1. ?disable all
2. ?disable new

After switching from all threads disabled to only new ones ``dm_disabled`` does not get updated to 1 (only new threads disabled).
image

Changes have been applied and tested.


martinbndr:
Should I add this to changelog or is it not very important for it to be in this changelog?

Taaku18 (approved):
Hello. I believe it was intentional for `?disable new` to not override `?disable all`. 
But I agree that's unintuitive, and your fix looks good.

PR #3279, opened 2023-04-12
Custom database name

### LINKED ISSUE #3191: Allow custom Mongo database names

Currently database name seems to be hardcoded in `clients.py`. It's a bit tedious to change `modmail_bot` strings in `clients.py` every time. This issue is especially apparent for those of us who run multiple modmail instances on the same machine, in our case it's different bot instances for different languages we support.

Perhaps add support for either database names in `CONNECTION_URI` in `.env` (mongodb+srv://mongodburi/**database_name**) (default to `modmail_bot` if none provided)

- [ ] Logviewer
- [ ] New config option

Alternatively make a new `.env` param such as `DATABASE_NAME`

### Who will this benefit?

People who run more than one instance of modmail bot on the same database, or people who prefer to have better configurability.

### Additional Information

_No response_

Hackerjef:
Allows for custom database name through mongo's connection URI

Closes #3191

### DIFF
--- a/core/clients.py
+++ b/core/clients.py
@@ -10,6 +10,7 @@ from discord.ext import commands
 from aiohttp import ClientResponseError, ClientResponse
 from motor.motor_asyncio import AsyncIOMotorClient
 from pymongo.errors import ConfigurationError
+from pymongo.uri_parser import parse_uri
 
 from core.models import InvalidConfigError, getLogger
 
@@ -445,7 +446,8 @@ class MongoDBClient(ApiClient):
                 raise RuntimeError
 
         try:
-            db = AsyncIOMotorClient(mongo_uri).modmail_bot
+            database = parse_uri(mongo_uri).get('database') if parse_uri(mongo_uri).get('database') else 'modmail_bot'
+            db = AsyncIOMotorClient(mongo_uri)[database]
         except ConfigurationError as e:
             logger.critical(
                 "Your MongoDB CONNECTION_URI might be copied wrong, try re-copying from the source again. "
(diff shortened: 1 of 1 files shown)


skybird23333:
I think `.env.example` should be updated as well

core/clients.py
```diff
@@ -445,7 +446,8 @@ def __init__(self, bot):
                 raise RuntimeError
 
         try:
-            db = AsyncIOMotorClient(mongo_uri).modmail_bot
+            database = parse_uri(mongo_uri).get('database') if parse_uri(mongo_uri).get('database') else 'modmail_bot'
```
raidensakura, line 449:
```suggestion
            database = parse_uri(mongo_uri).get('database') or 'modmail_bot'
```

core/clients.py
```diff
@@ -497,7 +499,8 @@ async def validate_database_connection(self, *, ssl_retry=True):
                     'run "Certificate.command" on MacOS, '
                     'and check certifi is up to date "pip3 install --upgrade certifi".'
                 )
-                self.db = AsyncIOMotorClient(mongo_uri, tlsAllowInvalidCertificates=True).modmail_bot
+                database = parse_uri(mongo_uri).get('database') if parse_uri(mongo_uri).get('database') else 'modmail_bot'
```
raidensakura, line 502:
```suggestion
                database = parse_uri(mongo_uri).get('database') or 'modmail_bot'
```

raidensakura (changes requested):
You can just use `or` in the assignment since it will ignore the first one if its a None value.

Jerrie-Aries:
What is the format of the URI if we want to use the custom db name?

raidensakura:
> What is the format of the URI if we want to use the custom db name?

`mongodb://myDBReader:D1fficultP%40ssw0rd@mongodb0.example.com:27017/your_db_name?someOptions`
as per [docs](https://www.mongodb.com/docs/manual/reference/connection-string/)

Hackerjef:
> I think `.env.example` should be updated as well
@skybird23333 I mean its still a valid mongodb uri so It should still be fine? /shrug

@raidensakura Thanks!

lorenzo132 (approved):

lorenzo132 (approved):

raidensakura (approved):

Vedinsoh:
Perhaps it would be good to update logviewer in the same way for consistency?

PR #3284, opened 2023-05-18
feat(build): drop root privileges, update .dockerignore

suprovsky:
This change will prevent from using root user in order to run the application. I also updated `.dockerignore` so you have less files in the image. I wasn't sure if you need `.git` for some reason in the image so I left it inside. But if you don't need it then it should be included in `.dockerignore` as well.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -138,13 +138,18 @@ temp/
 test.py
 
 # Other stuff
+.dockerignore
 .env.example
 .gitignore
 .github/
 app.json
 CHANGELOG.md
+docker-compose.yml
+LICENSE
+PRIVACY.md
 Procfile
 pyproject.toml
 README.md
 Pipfile
 Pipfile.lock
+SPONSORS.json
--- a/.dockerignore
+++ b/.dockerignore
@@ -144,6 +144,7 @@ test.py
 .github/
 app.json
 CHANGELOG.md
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -138,13 +138,19 @@ temp/
...
+Dockerfile
+docker-compose.yml
+LICENSE
+PRIVACY.md
 Procfile
 pyproject.toml
 README.md
 Pipfile
 Pipfile.lock
+SPONSORS.json
```
Taaku18, line 156:
```suggestion
.dockerignore
.env.example
.git/
.gitignore
.github/
app.json
CHANGELOG.md
Dockerfile
docker-compose.yml
Procfile
pyproject.toml
README.md
Pipfile
Pipfile.lock
```

Added `.git/` to the ignored list, as suggested.

Unignored `LICENSE`, `PRIVACY.md`, and `SPONSORS.json`. They do not affect the functionality of the bot, but due to their respective purposes, we prefer to include them in the image.

Taaku18 (approved):
Hello @suprovsky, thanks for the PR. 

I removed a few of the `.dockerignore` additions, since they should be included in the image. I also added the `.git` directory as suggested. 
I also slightly modified the Dockerfile by renaming the user to 'modmail' and moving to a higher layer.

PR #3285, opened 2023-05-24
Added Autoreact plugin

martinbndr:
This PR adds a autoreact plugin.
Commands:
?startreact
?stopreact
Requested by clay#6272 on discord.


StephenDaDev (approved):
Tested, works as designed. Looked through the code, all looks to be well. Icon is included, and fields are filled out. One small thing that I would say is you should probably disclose the 20 reaction limit in the help text for the command. I know it is sent in the activation message, but not in the help text. However, I would consider this to be an edge-case - I really doubt many will be affected by this-, so, consider it a nice to have and not a must have. 

So will still recommend adding to registry on next update @Taaku18

Taaku18 (dismissed):
Looks good to me!

PR #3292, opened 2023-07-13
Remove Heroku Reference from Debug Command

StephenDaDev:
Very few users are still using Heroku for hosting their instance of Modmail. As such, this comment could cause confusion in regard to where to find their logs. So, changing it to a general term, such as console, remedies this issue, and doesn't specifically target any one hosting method.


Taaku18 (changes requested):
Hi, you might've missed another reference lower down on L485. Other than that, looks good.

Taaku18 (approved):

PR #3300, opened 2023-08-12
Add an in

Kitty-Therapist:
Should be "This thread will close IN <time>" rather than "This thread will close <time>

This is a minor issue that has been missed for two years according to Lorenzo. I told them that I'd throw in a quick pull request for it.

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -448,7 +448,7 @@ class Modmail(commands.Cog):
 
         embed = discord.Embed(
             title="Scheduled close",
-            description=f"This thread will close {silent}{human_delta}.",
+            description=f"This thread will close in {silent}{human_delta}.",
             color=self.bot.error_color,
         )
 


cogs/modmail.py
```diff
@@ -448,7 +448,7 @@ async def send_scheduled_close_message(self, ctx, after, silent=False):
 
         embed = discord.Embed(
             title="Scheduled close",
-            description=f"This thread will close {silent}{human_delta}.",
+            description=f"This thread will close in {silent}{human_delta}.",
```
Jerrie-Aries, line 451: I think it should be put between `{silent}` and `{human_delta}`. Otherwise if it's silent, the output would be:
`This thread will close in silently 2 hours.`


```suggestion
            description=f"This thread will close {silent}in {human_delta}.",
```

Zinoviya99:
https://github.com/modmail-dev/Modmail/issues/3210

Already being worked on / reported.

raidensakura:
Duplicate of #3233, this is already fixed in develop branch.

Zinoviya99:
Already merged into the dev branch.

PR #3303, opened 2023-09-06
Add JSON logging support

### LINKED ISSUE #3246: JSON format log output

Using this bot with log aggregation services like Coralogix is currently less than optimal, because the log outputs are free text and not JSON. This means that, to identify log levels from log messages, we need to regex lines individually, which is brittle and prone to miscategorisation. Multiline log messages, e.g. stack traces, also get interpreted as separate messages, which further exacerbates the miscategorisation issue. For stack traces, this means that most of the stack trace gets incorrectly labelled as `info`, since it’s not realistically possible to tell a log aggregation service what is and is not part of the stack trace with just regex - or at least it isn't possible with any semblance of reasonable accuracy.

Furthermore, if one has some sort of conditional alert forwarding system, e.g. the one provided by Coralogix, preventing non-actionable issues reported as errors (such as the somewhat frequently recurring “ignoring exception in <something\>”) from triggering alerts is also difficult because of this. It can be done, but requires a lot of dancing around specific message wordings to avoid overexclusion and thus losing actually relevant alerts.

An optional, or preferably enabled-by-default in production deployments, JSON log output format.

- [ ] Logviewer
- [X] New config option

The current free text log output format, which works, but is brittle and suboptimal as detailed previously.

### Who will this benefit?

Any users who use log aggregation services, and anyone else who wishes to inspect their Modmail process logs in a structured manner.

### Additional Information

_No response_

nullishamy:
This adds support for JSON logging, along with the relevant options required.

This does not change the default log behaviour, so should be backwards compatible.

It is opt in via the {STREAM,FILE}_LOG_FORMAT option, which can be 'json' to use the new logger, or anything else to fallback to the old behaviour.

This is implemented in terms of a custom formatter, which is optionally applied to the log streams.

Closes #3246

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -53,7 +53,6 @@ from core.utils import extract_block_timestamp, normalize_alias, parse_alias, tr
 
 logger = getLogger(__name__)
 
-
 temp_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "temp")
 if not os.path.exists(temp_dir):
     os.mkdir(temp_dir)
@@ -71,7 +70,6 @@ class ModmailBot(commands.Bot):
         self.config.populate_cache()
 
         self.log_file_name = os.path.join(temp_dir, f"{self.token.split('.')[0]}.log")
-        self._configure_logging()
 
         intents = discord.Intents.all()
         if not self.config["enable_presence_intent"]:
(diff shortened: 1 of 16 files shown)


Jerrie-Aries:
Cool feature.

However, usually PRs should point to `development` branch. Plus we had quite a few of changes with logging in `development` (merged). Maybe you should take a look to avoid merge conflicts.

nullishamy:
That's done. Hope I understood what was going on with those commits in development, let me know if I changed / removed any code / functionality!

bot.py
```diff
@@ -43,6 +43,7 @@
     InvalidConfigError,
     PermissionLevel,
     SafeFormatter,
+    configure_log_format,
```
Jerrie-Aries, line 46: This is not used anywhere in this file, right?

bot.py
```diff
@@ -68,6 +69,8 @@ def __init__(self):
         self.config = ConfigManager(self)
         self.config.populate_cache()
 
+        self.log_file_name = os.path.join(temp_dir, f"{self.token.split('.')[0]}.log")
+
```
Jerrie-Aries, line 73: We no longer use this attribute. It has been removed and replaced with `self.log_file_path` (see line 92).

Jerrie-Aries (commented):
Just pointed up two things there.

Also if new config options were added (`stream_log_format`, `file_log_format`), it would be better if the help docs for those are also added in `core/config_help.json`.

Other than that, you need @Taaku18 to review this thoroughly.

bot.py
nullishamy: Thought I removed that, removed.

nullishamy:
Right, well that's those changes made. Can wait for Taku's review now.

PR #3304, opened 2023-09-08
Created confirm_thread_creation_timeout config

sebkuip:
This PR adds a config variable allowing users to set their own timeout for the confirm_thread_creation.

This required the addition of an `ints` category of configs for conversion as well.

To make the lower and upper limits work, I have added a new system allowing the setting of limits. This will be compared by a python statement like `lower <= value <= upper` only for keys specified. It will be tested when the value is set, and reverts to the previous value if invalid.

I have added a description in the json file, along with adding this config to the notes of all other related configs.


Taaku18:
Seems like there's a merge conflict. I don't have permission to make the necessary changes, but looks like it could be fixed by making `ConfirmThreadCreationView` in utils.py take a `timeout` parameter, and changing `view = ConfirmThreadCreationView(timeout=self.bot.config[confirm_thread_creation_timeout])`.

PR #3305, opened 2023-09-14
Add JSON logging support

### LINKED ISSUE #3246: JSON format log output

nullishamy:
This is a re-open of #3303, which got closed for some reason (not sure if the UI bugged, if I misclicked, or something else)

This adds support for JSON logging, along with the relevant options required.

This does not change the default log behaviour, so should be backwards compatible.

It is opt in via the {STREAM,FILE}_LOG_FORMAT option, which can be 'json' to use the new logger, or anything else to fallback to the old behaviour.

This is implemented in terms of a custom formatter, which is optionally applied to the log streams.

Closes #3246

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -138,11 +138,15 @@ temp/
 test.py
 
 # Other stuff
+.dockerignore
 .env.example
+.git/
 .gitignore
 .github/
 app.json
 CHANGELOG.md
+Dockerfile
+docker-compose.yml
 Procfile
 pyproject.toml
 README.md
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,5 @@
 TOKEN=MyBotToken
 LOG_URL=https://logviewername.herokuapp.com/
 GUILD_ID=1234567890
-MODMAIL_GUILD_ID=1234567890
 OWNERS=Owner1ID,Owner2ID,Owner3ID
 CONNECTION_URI=mongodb+srv://mongodburi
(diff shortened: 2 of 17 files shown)


nullishamy:
I actually don't know what the hell is going on here. I'm not closing these willfully. Going to report to GitHub if this happens again??

nullishamy:
Reporting this.

Taaku18 (approved):
Thanks for the commits, looks good!

PR #3306, opened 2023-09-29
Fix rate limit issue on raw reaction add/remove events.

Jerrie-Aries:
The rate limit on reaction events issue has been reported multiple times in support server. The latest one (as of this PR is created) can be found [here](https://discord.com/channels/1079074933008781362/1156308520686342276).
Currently the bot attempts to fetch the message using API call on every reaction made including in other than thread channels. It is quite unnecessary since we only need the message object to do thread stuff (mostly linking messages). More explanation of the issue can be found in support server's help channel [here](https://discord.com/channels/1079074933008781362/1086085790100094976/1096717438915194911).

This PR fixes the issue by making sure the thread exists before using API call to fetch the message.


Taaku18 (approved):
Looks good! Thanks for the PR.

PR #3317, opened 2023-11-26
Default autoupdate to disabled.

lorenzo132:
This feature has cause way to many issues for non experienced selfhosters. Its more profitable for everyone if this is disabled on default.


martinbndr:
May it would be better to change the behaiviour code wise, so an optional .env option to enable auto updates rather than to disable them. Also "updating" it like on the example file isnt that good as the changes wont affect the .env file so only new users configuring the bot first time having that change present.

PR #3333, opened 2024-02-20
Command: Avatar

martinbndr:
Adds an avatar command to the bot which makes it possible for owners to change the bots avatar directly via a command.


cogs/utility.py
```diff
@@ -2120,6 +2121,47 @@ def paginate(text: str):
...
+                title="Error",
+                description="Reading the attachment failed, is it valid?",
+                color=self.bot.error_color,
+            )
+            return await ctx.send(embed=embed)
+        try:
+            await self.bot.user.edit(avatar=dc_avatar)
+            logger.info("Bot Avatar updated.")
+        except Exception:
+            raise ValueError("Uploading the avatar to discord failed.")
```
raidensakura, line 2156: This doesn't look like it handles the exception gracefully, the command will just error silently giving no feedback to the user indicating the command has failed. Ideally this also should invoke `@trigger_typing` since it could take a few seconds to complete.

martinbndr:
Since discord added abillity to change this animated avatar/banner thing through the dev portal Im closing this. Guess users will find out during setup that they can set their avatar or banner through the portal.

PR #3334, opened 2024-02-26
Remove redundant `ctxs` initialization and `continue` bot.py

ya0guang:


lorenzo132:
Why it exists: alias expansion and menu “command” options can produce multiple steps (e.g., "a && b && c"). The code first builds all contexts (ctx objects) then iterates and invokes each.

We wont remove this as of now since its working functional. 

If its behaviour changes in the future we will change it to create-and-invoke inside the same loop.

PR #3339, opened 2024-04-10
Add logviewer into plugin registry

raidensakura:


sebkuip:
Raiden has long left the modmail community and has publicly stated to no longer support the plugins. I don't think having this PR open anymore is going to be beneficial to anyone

PR #3343, opened 2024-06-09
Alias add improvements

martinbndr:
- Fixes an issue with error message being incorrect.
- Protects eval command for being added into an alias if the user does not have permissions to it as Owner level etc. as the command is quite dangerous and should not be added by normal moderators (Could be used to nuke server for example).


StephenDaDev (dismissed):
Looks good.

cogs/utility.py
sebkuip, line 1282: This call to make_alias does not have the new ctx arg which would break editing aliases

sebkuip (changes requested):
There's an issue that you forgot to check causing errors.

martinbndr:
> There's an issue that you forgot to check causing errors.

Ih damn yea seems I forgot that, will fix it in next few days when I have time.

martinbndr:
Moved this to #3422. The reviewed issue has been implemented there. Closing this now

PR #3344, opened 2024-06-13
Plugin Registry: Support-Times

martinbndr:
Adds my Support-Times Plugin to the registry.
Feature: Makes you able to schedule enable/disable modmail at certain times with cron jobs.


lorenzo132:
Is this still a valid and functional plugin as of v4.2.1?

PR #3362, opened 2025-02-17
Allow using Discord threads

A5rocks:
Absolutely hacky support for Discord threads. There's a couple important details about Discord threads:
 - they do not support topics
 - (a bit less importantly, they can't be marked NSFW nor have overwrites)

The lack of topic support is troubling given that Modmail stuffs thread information into the channel topic. I updated some things to work based off the genesis message and -- since that does not contain the title -- disallowed setting titles for thread in Discord threads.

This can be used by setting `main_category_id` to a text channel, rather than a category channel. It worked in some basic tests I did locally, but I haven't checked in a more intensive environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ however, insignificant breaking changes do not guarantee a major version bump, s
 
 # v4.1.1
 
+### Breaking
+- Modmail threads are now potentially Discord threads
+
 ### Fixed
 - `?msglink` now supports threads with multiple recipients. ([PR #3341](https://github.com/modmail-dev/Modmail/pull/3341))
 - Fixed persistent notes not working due to discord.py internal change. ([PR #3324](https://github.com/modmail-dev/Modmail/pull/3324))
--- a/bot.py
+++ b/bot.py
@@ -305,13 +305,21 @@ class ModmailBot(commands.Bot):
             logger.debug("LOG_CHANNEL_ID was invalid, removed.")
             self.config.remove("log_channel_id")
         if self.main_category is not None:
-            try:
-                channel = self.main_category.channels[0]
-                self.config["log_channel_id"] = channel.id
-                logger.warning("No log channel set, setting #%s to be the log channel.", channel.name)
-                return channel
-            except IndexError:
-                pass
+            if isinstance(self.main_category, discord.CategoryChannel):
+                try:
+                    channel = self.main_category.channels[0]
+                    self.config["log_channel_id"] = channel.id
+                    logger.warning("No log channel set, setting #%s to be the log channel.", channel.name)
+                    return channel
+                except IndexError:
+                    pass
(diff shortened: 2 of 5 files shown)


martinbndr:
This would be a nice idea, but this completely overwrites the whole system to threads which can be a problem. There are servers that would like to use the old/original behaiviour from the current modmail. So to have a chance for being considered it would be better to implement this as an optional feature with a configuration variable.

lorenzo132:
> This would be a nice idea, but this completely overwrites the whole system to threads which can be a problem. There are servers that would like to use the old/original behaiviour from the current modmail. So to have a chance for being considered it would be better to implement this as an optional feature with a configuration variable.

I agree with this. If it were to be made a configurable option and thoroughly tested in an intensive environment, I’d be 100% okay with merging this.

A5rocks:
This still has the old behavior -- it only uses threads if `main_category_id` is a text channel.

martinbndr:
Ah okay

martinbndr:
I have not tested everything yet but a few things I have noticed:
- During the bot tries to create a modmail thread it´ll fails with this error:
```py
await channel.edit(topic=f"User ID: {user_id}")
TypeError: Thread.edit() got an unexpected keyword argument 'topic'```
```
- Bot logging shows:
`02/18/25 00:02:18 core.thread[1339] - WARNING: Found existing thread for 618805150756110336 but the channel is invalid.`
So there are these and may more parts that still does not fully recognize the main_category_d is a text channel

And may it would be better to let user configure it sth else. Because the variable name `main_categpory_id` would be a bit confusing for the users as they are thinking only categories are valid.

A5rocks:
I'm unable to reproduce either of those -- do you have some steps I should follow? (I can see where in the code they happen and could blindly put guards up, but I want to know in what case they happen)

martinbndr:
I just used the config set command and cofigured the main_category_id to a text channel.
adn then dmed the bot to start a new thread.

martinbndr (changes requested):
Line 1305/1322 (including your changes) could be the issue. Also wondering why you didnt get this error. Have you tested it properly?

A5rocks:
I've been testing repeating:
 1. Set main category id to a category
 2. Create a thread
 3. Close it
 4. Switch main category id to a text channel
 5. Create a thread
 6. Close it

Obviously with stuff like checking switching 3 and 4. Nothing gets logged or breaks.

I'll try looking at the code some more and see if I can see a way for maybe a .find call to race against sending a genesis message?

A5rocks:
I took a wild guess -- let me know if that fixes it (I am still unable to repro...)

sebkuip:
This PR is quite old, and would not work with a lot of new features. It's also not really of interest due to severe lack of features, and modmail now has a snooze features that might solve the issues that this PR tries to solve.

I'll be closing it for now.

PR #3364, opened 2025-02-28
Update recipient thread close

martinbndr:
- Updates the recipient_thread_close feature to use a button instead of a reaction.
- New config options: recipient_thread_close_button_label, recipient_thread_close_button_style
- The close_emoji default value is now None to make it possible changing the button to only have a label.

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -48,7 +48,15 @@ from core.models import (
 )
 from core.thread import ThreadManager
 from core.time import human_timedelta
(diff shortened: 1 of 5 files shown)


StephenDaDev (changes requested):
Please resolve merge conflicts

martinbndr:
Will do later, need to probably recreate it as it was made based on incorrect master branch instead it needs to be based on dev.
Also need to take a look at compatibility with the threadmenu

Issue #3372, opened 2025-05-02
Display Name


l1am2008:
### Is your feature request related to a problem? Please elaborate.

No

### Describe the solution you'd like

Ability to have the moderators server display name as their name on the reply.

### Does your solution involve any of the following?

- [ ] Logviewer
- [x] New config option

### Describe alternatives you've considered

Not really any alternatives.

### Who will this benefit?

Any servers that have a more professional naming scheme than the users discord handle.

lorenzo132:
Hi, 
This option has been added in #3392

PR #3373, opened 2025-05-03
Update thread.py

l1am2008:
Made it so that when you reply to a ticket, it uses your display name on the primary server instead of your discord username.

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -257,7 +257,8 @@ class Thread:
                 author = note["author"]
 
                 class Author:
-                    name = author["name"]
+                    print(author)
+                    name = author["nickname"]
                     id = author["id"]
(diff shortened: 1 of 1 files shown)


sebkuip:
This should be a config option. Not a forced change.

martinbndr:
Yea that's also my opinion. Making such a feature forced to change isn't good as many users still like to use the current behaviour. 
Because of that the user in the issue #3372 also suggested a new config option to avoid this. 

So in order to have this change considered for an update you need to make changes @LiamLuker

StephenDaDev:
I am going to close this as sepkuip mentioned, it does not have a descriptive title. And as both mentioned, we do not feel that this should be a forced change. Thank you.

PR #3378, opened 2025-06-01
Fix: Image url regex in thread send method

Zallom:
Revert image URL regex change in thread.py. The current regex only matches the protocol and domain, and no longer validates full image URLs.


obaseosa:
win

martinbndr:
Yea this fix needs to be implemented.

obaseosa:
Is this fixed yet?

Taaku18 (approved):

PR #3383, opened 2025-07-22
Feat: Renaming of snippets and aliases

martinbndr:
This PR adds two commands for renaming snippets and aliases for easier name editing.


lorenzo132 (approved):

StephenDaDev (approved):
This is good too. Thank you for the changes.

PR #3392, opened 2025-10-08
Internal bug fixes

lorenzo132:
fix: raceconditions, thread duplication on unsnooze, message queue for accuracy on highload

PR #3395, opened 2025-11-12
Allow ?logs [user] to be used outside of Modmail threads

SoAJeff:
This PR reverts a newly introduced (and likely unintended) change preventing `?logs [user]` from being used outside of a Modmail thread.

When the logs command is used with a user argument, the current version of Modmail requires the command to be run in a Modmail thread. This behavior change appears to have been introduced with commit 9d4b8e3 as previous versions allowed the command to be used outside of Modmail threads.

The other logs commands (closed-by, delete, key, responded, search) do not require the command to be run in a thread.  These commands should all be consistent with one another.  From a usability standpoint, having to open a Modmail thread to review a user's logs (even if not intending to send the user a message) seems counterintuitive.


lorenzo132:
Hi, thanks for the PR. 
I have merged this pr into #3392

Issue #3357, opened 2025-01-16
Clarify why the update command fails on non git installs


sebkuip:
### Is your feature request related to a problem? Please elaborate.

When a user runs the ?update command, the bot will run `git pull` and then resetart itself. This is nice unless the bot isn't installed by git, which a lot of bots are not. The bot just spits out the raw error of not finding a .git folder which is confusing to most user.

### Describe the solution you'd like

I would suggest that the ?update command first checks the presence of a .git folder (and anything else required for update to work) before starting. If this isn't found, the bot should spit out a clear message to the user instructing that their hosting platform does not support automatic updates, and instructions on how to manually update.

### Does your solution involve any of the following?

- [ ] Logviewer
- [ ] New config option

### Describe alternatives you've considered

N/A

### Who will this benefit?

This will mostly benefit users that are less known with git and programming as they will get more clear errors and instructions on what to do. It will also benefit support staff as they have to explain this less.

martinbndr:
Im thinking how the message could be written that the user gets when ther´s no git repo
Thoulgt of this embed:
```py
embed = discord.Embed(
    title="Update Command Unavailable",
    description="The bot cannot be updated due to not being installed via a git."
                        "You need to manually update the bot according to your hosting method.",
    color=discord.Color.red()
    )```
Or any other suggestions?

lorenzo132:
This has been added in https://github.com/modmail-dev/Modmail/pull/3405, ill be closing this issue.

PR #3358, opened 2025-01-16
Check for local git repository

martinbndr:
Implements a checking mechanismen for local git repository. Requested in #3357 which checks for a valid git repository that the .git folder exsisting and checks for some typical git files and folders to validate that its most likley working with an update.

PR #3405, opened 2025-12-01
Git repository checking

martinbndr:
Implements a checking mechanismen for local git repository. Requested in #3357 which checks for a valid git repository that the .git folder exsisting and checks for some typical git files and folders to validate that its most likley working with an update.

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -794,6 +794,33 @@ class ModmailBot(commands.Bot):
         logger.debug("User blocked, user %s.", author.name)
         return False
 
+    def check_local_git(self) -> bool:
+        """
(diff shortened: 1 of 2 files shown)


martinbndr:
I thought a new pr in a new branch would fix it. Because I messed up in the old one #3358.
@Taaku18 So when implementing changes the branches should be created based on master? That was my last info or (thats why the "branch out of date" shows up should we create it based of development?

martinbndr:
Nvm will discuss through discord. Will probably redo this via an own branch on modmail repo instead of my fork

PR #3410, opened 2025-12-01
Update dev dependencies

martinbndr:
Updates some dev dependencies to a more updated version. Recently there were issues with black that should be resolved by a newer version. I also guess its not a problem to update the pylint module. Overall we should maintain dependencies a bit more in future as some libs also include bug fixes etc.


martinbndr:
Will be retried later

PR #3412, opened 2025-12-04
Fix: Blackformatting, workflows. Add snooze/unsnooze events.

lorenzo132:


Pipfile
sebkuip, line 1: Has this dependency change been verified to work? Needs to go through some longer testing to ensure it all runs smoothly without weird new errors.

Also requirements.txt is not updated.

sebkuip (changes requested):
Most changes are fine, but the dependency changes need some fixing up.

Also next time make separate PRs for different changes.

sebkuip (approved):
LGTM

Pipfile
```diff
@@ -7,28 +7,29 @@ verify_ssl = true
 bandit = ">=1.7.5"
 black = "==23.11.0"
 pylint = "==3.0.2"
-tomli = "==2.2.1"  # Needed for black on Python < 3.11
+tomli = "==2.2.1"
 
 [packages]
 aiohttp = "==3.13.2"
-async-timeout = {version = "==5.0.1", markers = "python_version < '3.11'"}  # Required by aiohttp
```
StephenDaDev, line 14: I feel like we should not remove the comments explaining why it is included? Is there anything to support this idea?

lorenzo132: > Is there anything to support this idea?

What exactly do you mean by this?

StephenDaDev: Is there some convention or reason to support removing comments from the requirements list

lorenzo132: No reason, no. But since other packages also do not have those, i want to put everything on 1 line.

StephenDaDev (approved):

martinbndr (approved):

PR #3414, opened 2025-12-05
fix: show forwarded message in logviewer.

### LINKED ISSUE #2876: Support editing and deleting plain messages

### LINKED ISSUE #3420: Attachment support for forwarded messages

Currently if a message is forwarded, the bot will show the message content. However, images are not shown. If a forward is only an image it will show “No content” like the image below.

Image

Forward images as well

- [ ] Logviewer
- [ ] New config option

No alternative

### Who will this benefit?

Staff using modmail in a situation where they could receive images.

### Additional Information

_No response_

lorenzo132:
Forwarded message now will show correctly in logviewer.
image


StephenDaDev (dismissed):
LGTM.

core/utils.py
```diff
@@ -658,39 +658,49 @@ def extract_forwarded_content(message) -> typing.Optional[str]:
...
+                if getattr(snap, "attachments", None):
+                    attachment_info = ", ".join(
+                        [getattr(a, "filename", "Unknown") for a in snap.attachments[:3]]
+                    )
+                    if len(snap.attachments) > 3:
+                        attachment_info += f" (+{len(snap.attachments) - 3} more)"
+                    formatted_part += f"[Attachments: {attachment_info}]\n"
+
+                # Add source link to the container message since snapshot doesn't have its own public link
+                formatted_part += f"\n**Source:** {message.jump_url}"
```
SoAJeff: This always returns a link to the forwarded message instance that is in the DM with the bot, rather than to the original message that was forwarded - which won't be accessible to the moderator reviewing the content in the logviewer.

The "Context" line when a forwarded message is sent in a Modmail ticket channel (which appears when the forwarded message location is one that the bot is able to fetch) would be a better solution here - perhaps something like what is done in [thread.py](https://github.com/modmail-dev/Modmail/blob/6e75c448f24390bfd27990fbd33729e9295099cf/core/thread.py#L1855)

SoAJeff (commented):
Appreciate the quick turn-around on getting a fix for this.  As a mod in a server that receives a lot of forwarded messages in Modmail threads to us, the recent additions in supporting forwarded messages both on the bot side and the logviewer side have been very helpful to us.

Just one minor correction I saw to improve usability for mods reviewing Modmail logs after the fact.

core/utils.py
lorenzo132: That would be a nice solution yea. If you are able to PR this that would be great, otherwise i will look into this either this week or next week.

StephenDaDev (dismissed):

core/utils.py
```diff
@@ -640,6 +641,37 @@ def __init__(self):
...
+    ----------
+    message : discord.Message
+        The message to extract attachments from.
+
+    Returns
+    -------
+    List[Tuple[str, str]]
+        List of (url, filename) tuples for attachments.
+    """
+    import discord
```
sebkuip: Why do we have an import in the middle of a function? Move this to the top of the file.

core/utils.py
```diff
@@ -640,6 +641,37 @@ def __init__(self):
...
+
+    attachments = []
+    try:
+        if getattr(message, "message_snapshots", None):
```
sebkuip: Using hasattr() might be a cleaner solution for these checks. Internally python uses getattr() to make it work, but it's better for readability.
https://docs.python.org/3/library/functions.html#hasattr

Please also fix the rest of these instances (I won't be commenting on them all.)

core/utils.py
```diff
@@ -640,6 +641,37 @@ def __init__(self):
...
+            for snap in message.message_snapshots:
+                if getattr(snap, "attachments", None):
+                    for a in snap.attachments:
+                        url = getattr(a, "url", None)
+                        filename = getattr(a, "filename", "Unknown")
+                        if url:
+                            attachments.append((url.split("?")[0], filename))
```
sebkuip, line 667: Not sure if it's because of this PR, but forwarded attachments show up incorrectly formatted in the logviewer. They show up as `[image](link)` and don't properly embed. An image sent directly does show up properly.

Image

sebkuip (changes requested):

core/utils.py
lorenzo132: I have been able to partialy resolve this request. Images would show as following for some reason.
image

core/utils.py
lorenzo132: Resolved.

martinbndr (dismissed):

core/clients.py
```diff
@@ -661,9 +661,18 @@ async def append_log(
         channel_id: str = "",
         type_: str = "thread_message",
     ) -> dict:
+        from core.utils import extract_forwarded_content
```
sebkuip: Import needs to be moved to the top of the file. I don't really see any real reason to have this inside the method here as it's not an expensive import that would slow down startup time significantly.

sebkuip (changes requested):
I've tested the message forwarding locally and it seems to work just fine. I find one issue with the code regarding the placement of an import. For the rest it looks all good. If you can get this one change fixed up it should be all good to merge.

sebkuip (approved):
LGTM now. All thing pointed out have been resolved.

StephenDaDev (approved):
Looks good, tested working.

martinbndr (approved):

PR #3418, opened 2025-12-13
arged responses

lorenzo132:
This add a args feature, for example if your arg is called `argreply`

You can do: `?reply Hello, read the following: {argreply}

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -16,7 +16,7 @@ from discord.utils import escape_markdown
 from dateutil import parser
 
 from core import checks
-from core.models import DMDisabled, PermissionLevel, SimilarCategoryConverter, getLogger
+from core.models import DMDisabled, PermissionLevel, SimilarCategoryConverter, UnseenFormatter, getLogger
 from core.paginator import EmbedPaginatorSession
 from core.thread import Thread
 from core.time import UserFriendlyTime, human_timedelta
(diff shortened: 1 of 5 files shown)


cogs/modmail.py
```diff
@@ -535,6 +535,199 @@ async def snippet_rename(self, ctx, name: str.lower, *, value):
...
+            val = truncate(escape_code_block(self.bot.args[name]), 2048 - 7)
+            embed = discord.Embed(
+                title=f'Raw arg - "{name}":',
+                description=f"```\n{val}```",
+                color=self.bot.main_color,
+            )
+
+        return await ctx.send(embed=embed)
+
+    @args.command(name="add", aliases=["create", "make"])
```
StephenDaDev, line 623: Should we begin to add "new" as another alias for these commands (across aliases/snippets/args)? I know it was brought up a few times over the past few years and never went anywhere.

cogs/modmail.py
```diff
@@ -535,6 +535,199 @@ async def snippet_rename(self, ctx, name: str.lower, *, value):
...
+            title="Added arg",
+            color=self.bot.main_color,
+            description="Successfully created arg.",
+        )
+        return await ctx.send(embed=embed)
+
+    @args.command(name="remove", aliases=["del", "delete"])
+    @checks.has_permissions(PermissionLevel.SUPPORTER)
+    async def args_remove(self, ctx, *, name: str.lower):
+        """Remove an arg."""
```
StephenDaDev, line 671: Also, are we calling this feature "args" or "arguments". If we decide to stick to "args" 24/7, this is fine. If we choose to have it "arguments" but shortened to "args" we should update this to say `Remove an argument.` instead. (This goes for all command help items in this PR).

martinbndr (dismissed):

StephenDaDev (dismissed):

cogs/modmail.py
lorenzo132: I personally do not think we should. I'd rather stick the the current flow.

cogs/modmail.py
lorenzo132: Wr sticking to arg. This is because it is easier to use(shorter) and makes it user friendly.

StephenDaDev (dismissed):

martinbndr (dismissed):

StephenDaDev (dismissed):

martinbndr (approved):

StephenDaDev (approved):

cogs/modmail.py
```diff
@@ -1510,6 +1699,9 @@ async def reply(self, ctx, *, msg: str = ""):
         automatically embedding image URLs.
         """
 
+        if self.bot.args:
+            msg = UnseenFormatter().format(msg, **self.bot.args)
```
sebkuip, line 1726: There is no check for total message length after applying args. If I make an arg that's very long (I tested with a long part of lorem ipsum) then make a long reply plus the arg going over the limit, I just get an "unknown error". The bot logs show it goes over the limit.
As per the docs (screenshot below), the total length of an embed cannot exceed 6000 characters, and the descritpion may only be 4096 characters long. Please add a check to give a human error.

Image

sebkuip: Here is the result of a long message
Image

Image

cogs/modmail.py
```diff
@@ -535,6 +535,195 @@ async def snippet_rename(self, ctx, name: str.lower, *, value):
...
+            self.bot.args[value] = old_arg_value
+            await self.bot.config.update()
+
+            embed = discord.Embed(
+                title="Renamed arg",
+                color=self.bot.main_color,
+                description=f'`{name}` has been renamed to "{value}".',
+            )
+        else:
+            embed = create_not_found_embed(name, self.bot.args.keys(), "Arg")
```
sebkuip: For readability, this might be better fit as a guard clause at the start of the function. So
```
if name not in self.bot.args:
    return await ctx.send(embed=create_not_found_embed(name, self.bot.args.keys(), "Arg"))
```
or alternatively
```
if name not in self.bot.args:
    embed = create_not_found_embed(name, self.bot.args.keys(), "Arg")
```

cogs/modmail.py
```diff
@@ -535,6 +535,195 @@ async def snippet_rename(self, ctx, name: str.lower, *, value):
...
+    @checks.has_permissions(PermissionLevel.SUPPORTER)
+    async def args_add(self, ctx, name: str.lower, *, value: commands.clean_content):
+        """
+        Add an arg.
+
+        Simply to add an arg, do: ```
+        {prefix}args add name value
+        ```
+        """
+        if name in self.bot.args:
```
sebkuip, line 633: This should also check that there is no conflict with the keywords from formatreply (channel, recipient and author). Currently this causes a silent fail without error. See screenshots.

Image

Image

cogs/modmail.py
sebkuip, line 1726: Replace this formatter to be consistent with how formatreply works (using `self.bot.formater.format()`)
This is done correctly in the formatreply command.

cogs/utility.py
```diff
@@ -109,7 +109,7 @@ async def format_cog_help(self, cog, *, no_cog=False):
         return embeds
 
     def process_help_msg(self, help_: str):
-        return help_.format(prefix=self.context.clean_prefix) if help_ else "No help message."
+        return help_.replace("{prefix}", self.context.clean_prefix) if help_ else "No help message."
```
sebkuip: Why is this being changed out? As far as I'm aware, the format() call was working just fine. If really needed, please elaborate why this is changed out.

sebkuip (changes requested):
There are some issues in this PR that need to be addressed beforehand.

StephenDaDev:
I discussed this with Lorenzo and removed this from this update to allow additional time for corrections while still working towards the important changes in 4.3.0

martinbndr (dismissed):
Looks good to me

StephenDaDev (dismissed):
seems fine w/ changes

cogs/modmail.py
sebkuip: Small typo here, pointed out by copilot. "don't" is missing it's apostrophe

sebkuip: I would change this text to `See"{self.bot.prefix}help args add" for how to add an arg.`

sebkuip: Needs a comma after "simply"

cogs/modmail.py
```diff
@@ -1700,7 +1723,17 @@ async def reply(self, ctx, *, msg: str = ""):
         """
 
         if self.bot.args:
-            msg = UnseenFormatter().format(msg, **self.bot.args)
+            msg = self.bot.formatter.format(msg, **self.bot.args)
```
sebkuip: The system breaks when an arg with a `.` character in the name is added. It tries to resolve it as an attribute causing either an `<invalid>` or an error to appear.

Using a string as an arg causes no reply and a silent error.
Image
Image

Using a forbidden arg causes an `<invalid>` to appear (here I did `?r testing {channel.name}`)
Image

Formatreply correctly works and doesn't cause issues

sebkuip (changes requested):
I found some bugs, and some typos. Please have these fixed. Thanks!

cogs/modmail.py
```diff
@@ -1532,10 +1757,22 @@ async def freply(self, ctx, *, msg: str = ""):
         """
         msg = self.bot.formatter.format(
             msg,
+            **self.bot.args,
```
sebkuip, line 1760: Whilst the issue has been fixed in the regular reply command, in freply (and i assume similar will happen with afreply) it still glitches out when you use an arg with a dot in the name, as it tries to process it like a regular arg as functionality from freply.

https://youtu.be/RiffXdUfI7g

sebkuip (changes requested):
Found a bug with dotted names within freply (and afreply likely as well)

cogs/modmail.py
lorenzo132: This should now be resolved as the reply commands share a safe formatter

sebkuip (approved):
I think i found everything I could think off. Looks good to me now.

cogs/modmail.py
```diff
@@ -1510,6 +1722,19 @@ async def reply(self, ctx, *, msg: str = ""):
         automatically embedding image URLs.
         """
 
+        if self.bot.args:
+            msg = UnseenFormatter().format(msg, **self.bot.args)
```
sebkuip, line 1726: As github copilot noted: Having args causes unmatched braces to throw errors.

Image

Image

sebkuip (changes requested):
Ignore previous review. Found a bug with copilots help

cogs/modmail.py
lorenzo132: This is resolved.

cogs/modmail.py
```diff
@@ -535,6 +539,214 @@ async def snippet_rename(self, ctx, name: str.lower, *, value):
...
+                color=self.bot.main_color,
+                description=f'`{name}` will now be replaced with "{value}".',
+            )
+        else:
+            embed = create_not_found_embed(name, self.bot.args.keys(), "Arg")
+        await ctx.send(embed=embed)
+
+    @args.command(name="rename")
+    @checks.has_permissions(PermissionLevel.SUPPORTER)
+    async def args_rename(self, ctx, name: str.lower, *, value):
```
sebkuip: The value is missing the `: commands.clean_content` converter

cogs/modmail.py
```diff
@@ -535,6 +539,214 @@ async def snippet_rename(self, ctx, name: str.lower, *, value):
...
+                description="You don't have any args at the moment.",
+            )
+            embed.set_footer(text=f'See "{self.bot.prefix}help args add" for how to add an arg.')
+            embed.set_author(
+                name="Args",
+                icon_url=self.bot.get_guild_icon(guild=ctx.guild, size=128),
+            )
+            return await ctx.send(embed=embed)
+
+        embeds = [discord.Embed(color=self.bot.main_color) for _ in range((len(self.bot.args) // 10) + 1)]
```
sebkuip: As stated by copilot, having exactly 10 args causes an empty args page

Image

sebkuip (changes requested):
Small things from copilot's suggestions that are confirmed bugs

sebkuip (approved):
Look all good now. Can't find further bugs myself

StephenDaDev (approved):

martinbndr (approved):

PR #3421, opened 2025-12-18
feat: adding attachments to snippets.

lorenzo132:
This adds the ability to add attachments to snippets.

### DIFF
--- a/cogs/modmail.py
+++ b/cogs/modmail.py
@@ -112,6 +112,30 @@ class Modmail(commands.Cog):
             return int(match.group(1))
         return None
 
+    def _get_snippet_text(self, snippet_data) -> str:
+        """
+        Extract text from a snippet, handling both old string format and new dict format.
+
+        Parameters
+        ----------
+        snippet_data : str or dict
+            The snippet data (either old string format or new dict format).
+
+        Returns
+        -------
+        str
+            The text content of the snippet.
+        """
+        if isinstance(snippet_data, str):
+            return snippet_data
+        elif isinstance(snippet_data, dict):
+            return snippet_data.get("text", "")
+        return ""
+
+    def _has_snippet_attachment(self, snippet_data) -> bool:
+        """Check if a snippet has an attachment."""
+        return isinstance(snippet_data, dict) and bool(snippet_data.get("file_id"))
+
     @commands.command()
     @trigger_typing
     @checks.has_permissions(PermissionLevel.OWNER)
@@ -246,10 +270,17 @@ class Modmail(commands.Cog):
             if snippet_name is None:
                 embed = create_not_found_embed(name, self.bot.snippets.keys(), "Snippet")
             else:
-                val = self.bot.snippets[snippet_name]
+                snippet_data = self.bot.snippets[snippet_name]
+                snippet_text = self._get_snippet_text(snippet_data)
+                has_attachment = self._has_snippet_attachment(snippet_data)
+
+                description = snippet_text if snippet_text else "(No text content)"
+                if has_attachment:
+                    description += "\n\n📎 *This snippet has an attachment.*"
+
                 embed = discord.Embed(
                     title=f'Snippet - "{snippet_name}":',
-                    description=val,
+                    description=description,
                     color=self.bot.main_color,
                 )
             return await ctx.send(embed=embed)
@@ -270,10 +301,15 @@ class Modmail(commands.Cog):
         for embed in embeds:
             embed.set_author(name="Snippets", icon_url=self.bot.get_guild_icon(guild=ctx.guild, size=128))
 
-        for i, snippet in enumerate(sorted(self.bot.snippets.items())):
-            embeds[i // 10].add_field(
-                name=snippet[0], value=return_or_truncate(snippet[1], 350), inline=False
-            )
+        for i, (snippet_name, snippet_data) in enumerate(sorted(self.bot.snippets.items())):
+            snippet_text = self._get_snippet_text(snippet_data)
+            has_attachment = self._has_snippet_attachment(snippet_data)
+
+            display_value = return_or_truncate(snippet_text, 350) if snippet_text else "(No text)"
+            if has_attachment:
+                display_value = "📎 " + display_value
+
+            embeds[i // 10].add_field(name=snippet_name, value=display_value, inline=False)
 
         session = EmbedPaginatorSession(ctx, *embeds)
         await session.run()
@@ -288,10 +324,18 @@ class Modmail(commands.Cog):
         if snippet_name is None:
             embed = create_not_found_embed(name, self.bot.snippets.keys(), "Snippet")
         else:
-            val = truncate(escape_code_block(self.bot.snippets[snippet_name]), 2048 - 7)
+            snippet_data = self.bot.snippets[snippet_name]
+            snippet_text = self._get_snippet_text(snippet_data)
+            has_attachment = self._has_snippet_attachment(snippet_data)
+
+            val = truncate(escape_code_block(snippet_text), 2048 - 7) if snippet_text else "(No text content)"
+            description = f"```\n{val}```"
+            if has_attachment:
+                description += "\n\n📎 *This snippet has an attachment.*"
+
             embed = discord.Embed(
                 title=f'Raw snippet - "{snippet_name}":',
-                description=f"```\n{val}```",
+                description=description,
                 color=self.bot.main_color,
             )
 
@@ -299,9 +343,9 @@ class Modmail(commands.Cog):
 
     @snippet.command(name="add", aliases=["create", "make"])
     @checks.has_permissions(PermissionLevel.SUPPORTER)
-    async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_content):
+    async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_content = None):
         """
-        Add a snippet.
+        Add a snippet with an optional attachment.
 
         Simply to add a snippet, do: ```
         {prefix}snippet add hey hello there :)
@@ -311,6 +355,8 @@ class Modmail(commands.Cog):
         To add a multi-word snippet name, use quotes: ```
         {prefix}snippet add "two word" this is a two word snippet.
         ```
+
+        You can also attach a file (max 10 MB) to include with the snippet.
         """
         if self.bot.get_command(name):
             embed = discord.Embed(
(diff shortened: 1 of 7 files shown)


cogs/modmail.py
```diff
@@ -343,14 +389,122 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                            "Only the command author can cancel.", ephemeral=True
+                        )
+                    confirmed = False
+                    await interaction.response.edit_message(
+                        content="❌ Cancelled. Snippet not created.", view=None, embed=None
+                    )
+                    view.stop()
+
+                confirm_button = discord.ui.Button(label="✓ Confirm", style=discord.ButtonStyle.green)
+                cancel_button = discord.ui.Button(label="✗ Cancel", style=discord.ButtonStyle.red)
```
StephenDaDev: Are these symbols consistent with the confirm and cancel across the bot? I feel like usually it is an emoji or just no symbol.

cogs/modmail.py
```diff
@@ -343,14 +389,122 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                confirm_button.callback = confirm_callback
+                cancel_button.callback = cancel_callback
+                view.add_item(confirm_button)
+                view.add_item(cancel_button)
+
+                embed = discord.Embed(
+                    title="Confirm Large Attachment",
+                    description=f"The attachment is {attachment.size / (1024 * 1024):.2f} MB (≥2 MB).\n"
```
StephenDaDev: the 1024 1024 makes this a MiB not MB.

StephenDaDev:
I have began a review which will result in a "changes requested" verdict. I will need additional time to complete a full review. I hope to have it completed by sometime around 12PM tomorrow, Eastern.

martinbndr (dismissed):
image

Is it intented, that an image is sent seperate comparing to the normal reply? If not maybe we should consider letting images being added to the embed aswell to have it consistent to the replying?

lorenzo132:
> image
> Is it intented, that an image is sent seperate comparing to the normal reply? If not maybe we should consider letting images being added to the embed aswell to have it consistent to the replying?



I have changed the behavior, now it just sets the attachment.(I was half asleep when i made this lolololol)

StephenDaDev (approved):
tested working, no serious issues found

cogs/modmail.py
```diff
@@ -250,10 +274,17 @@ async def snippet(self, ctx, *, name: str.lower = None):
             if snippet_name is None:
                 embed = create_not_found_embed(name, self.bot.snippets.keys(), "Snippet")
             else:
-                val = self.bot.snippets[snippet_name]
+                snippet_data = self.bot.snippets[snippet_name]
+                snippet_text = self._get_snippet_text(snippet_data)
+                has_attachment = self._has_snippet_attachment(snippet_data)
+
+                description = snippet_text if snippet_text else "(No text content)"
```
sebkuip: If a snippet is only an attachment, having it show up as `(No text content) This snippet has an attachment` seems weird. I would put it as something like `this is an attachment-only snippet`

cogs/modmail.py
```diff
@@ -274,10 +305,15 @@ async def snippet(self, ctx, *, name: str.lower = None):
...
-            embeds[i // 10].add_field(
-                name=snippet[0], value=return_or_truncate(snippet[1], 350), inline=False
-            )
+        for i, (snippet_name, snippet_data) in enumerate(sorted(self.bot.snippets.items())):
+            snippet_text = self._get_snippet_text(snippet_data)
+            has_attachment = self._has_snippet_attachment(snippet_data)
+
+            display_value = return_or_truncate(snippet_text, 350) if snippet_text else "(No text)"
+            if has_attachment:
+                display_value = "📎 " + display_value
```
sebkuip, line 454: Maybe make the embed have a footer text explaining that the paperclip emoji means the embed has an attachment?

cogs/modmail.py
```diff
@@ -250,10 +274,17 @@ async def snippet(self, ctx, *, name: str.lower = None):
...
                 embed = create_not_found_embed(name, self.bot.snippets.keys(), "Snippet")
             else:
-                val = self.bot.snippets[snippet_name]
+                snippet_data = self.bot.snippets[snippet_name]
+                snippet_text = self._get_snippet_text(snippet_data)
+                has_attachment = self._has_snippet_attachment(snippet_data)
+
+                description = snippet_text if snippet_text else "(No text content)"
+                if has_attachment:
+                    description += "\n\n📎 *This snippet has an attachment.*"
```
sebkuip: I'd say show the attachment when you request the snippet. In the list it's fine to leave out but if you just pull up that snippet it should show.

cogs/modmail.py
```diff
@@ -292,20 +328,28 @@ async def snippet_raw(self, ctx, *, name: str.lower):
...
         else:
-            val = truncate(escape_code_block(self.bot.snippets[snippet_name]), 2048 - 7)
+            snippet_data = self.bot.snippets[snippet_name]
+            snippet_text = self._get_snippet_text(snippet_data)
+            has_attachment = self._has_snippet_attachment(snippet_data)
+
+            val = truncate(escape_code_block(snippet_text), 2048 - 7) if snippet_text else "(No text content)"
+            description = f"```\n{val}```"
+            if has_attachment:
+                description += "\n\n📎 *This snippet has an attachment.*"
```
sebkuip: Similar to the regular version of this command, show the attachment so the staff member can see what it sends.

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
             )
             return await ctx.send(embed=embed)
 
-        self.bot.snippets[name] = value
-        await self.bot.config.update()
+        # Handle optional attachment
+        file_id = None
+        attachment_info = None
+        if ctx.message.attachments:
+            attachment = ctx.message.attachments[0]
```
sebkuip, line 562: Don't silently ignore multiple attachments. Either allow it, or give the user an error that only a single attachment is allowed.

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+
+            # Validate file size
+            max_size_mb = self.bot.config.get("snippet_attachment_max_size")
+            max_size_bytes = max_size_mb * 1024 * 1024
```
sebkuip: Wouldn't `max_size_mb * 1024 ** 2` be cleaner? Possibly `max_size_mb * (1024 ** 2)` to make it more clear?
Also add a short comment to explain why this calculation is done.

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                    title="Error",
+                    color=self.bot.error_color,
+                    description=f"Attachment exceeds the maximum file size of {max_size_mb} MB. "
+                    f"Your file is {attachment.size / (1024 * 1024):.2f} MB.",
+                )
+                return await ctx.send(embed=embed)
+
+            # Confirmation for attachments 2MB or higher
+            confirm_msg = None
+            if attachment.size >= 2 * 1024 * 1024:
```
sebkuip: Similar as comment above

martinbndr (dismissed):
Looks good.

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                logger.error("Failed to upload snippet attachment: %s", e)
+                embed = discord.Embed(
+                    title="Error",
+                    color=self.bot.error_color,
+                    description="Failed to upload attachment. Please try again.",
+                )
+                return await ctx.send(embed=embed)
+
+        # Require at least text or attachment
+        if not value and not file_id:
```
sebkuip: This if statement acts as a guard, and should be at the top of the method. Probably change it to `if not value and len(message.attachments) == 0:`

cogs/modmail.py
```diff
@@ -457,28 +620,114 @@ async def snippet_remove(self, ctx, *, name: str.lower):
             )
             self.bot.snippets.pop(name)
             await self.bot.config.update()
+            if file_id and not await self.bot.api.delete_snippet_attachment(file_id):
+                logger.warning("Failed to delete snippet attachment for %s.", name)
```
sebkuip: This should either inform the user in chat, or fail the command entirely so you don't end up with tons of data on gridFS when it errors a few times.

cogs/modmail.py
```diff
@@ -457,28 +620,114 @@ async def snippet_remove(self, ctx, *, name: str.lower):
...
         """
-        Edit a snippet.
+        Edit a snippet's text and/or attachment.
 
         To edit a multi-word snippet name, use quotes: ```
         {prefix}snippet edit "two word" this is a new two word snippet.
         ```
+
+        Attach a new file to replace the existing attachment.
+        Provide text without attachment to keep the existing attachment.
```
sebkuip: I feel like an edit should be able to remove the attachment as well. Isn't it better that providing no attachment deletes it, instead of having the user needing to delete the snippet if they want the attachment gone?

cogs/utility.py
```diff
@@ -109,7 +109,7 @@ async def format_cog_help(self, cog, *, no_cog=False):
         return embeds
 
     def process_help_msg(self, help_: str):
-        return help_.format(prefix=self.context.clean_prefix) if help_ else "No help message."
+        return help_.replace("{prefix}", self.context.clean_prefix) if help_ else "No help message."
```
sebkuip: Why is this randomly edited? This is out of scope for the PR

core/config.py
```diff
@@ -422,6 +426,14 @@ def get(self, key: str, *, convert: bool = True) -> typing.Any:
                     logger.warning("Invalid %s %s.", key, value)
                     value = self.remove(key)
 
+        elif key in self.megabytes:
+            if not isinstance(value, int):
+                try:
+                    value = int(value)
```
sebkuip: Might also want to guard against negative values. Also this causes rounding when a comma value is entered, is that intentional?

sebkuip (dismissed):
Review for now. Did not check all, and is purely a code review. Have to do some testing later myself and look further into the code.

cogs/modmail.py
```diff
@@ -315,6 +359,8 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
         To add a multi-word snippet name, use quotes: ```
         {prefix}snippet add "two word" this is a two word snippet.
         ```
+
+        You can also attach a file (max 10 MB) to include with the snippet.
```
StephenDaDev: I think this needs to have a disclaimer that explains that this is storage intensive and may cause issues if used frequently.

cogs/modmail.py
StephenDaDev: on L363 you are informing users that you have a 10MB limit but you are calculating in MiB instead

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+            attachment = ctx.message.attachments[0]
+
+            # Validate file size
+            max_size_mb = self.bot.config.get("snippet_attachment_max_size")
+            max_size_bytes = max_size_mb * 1024 * 1024
+            if attachment.size > max_size_bytes:
+                embed = discord.Embed(
+                    title="Error",
+                    color=self.bot.error_color,
+                    description=f"Attachment exceeds the maximum file size of {max_size_mb} MB. "
```
StephenDaDev: same as above

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+
+                async def cancel_callback(interaction: discord.Interaction):
+                    nonlocal confirmed
+                    if interaction.user.id != ctx.author.id:
+                        return await interaction.response.send_message(
+                            "Only the command author can cancel.", ephemeral=True
+                        )
+                    confirmed = False
+                    await interaction.response.edit_message(
+                        content="❌ Cancelled. Snippet not created.", view=None, embed=None
```
StephenDaDev: I think the use of emojis does not align with informal styling guidelines of the bot, I can't think of any other cases where they are used in this context

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                    )
+                    view.stop()
+
+                confirm_button = discord.ui.Button(label="✓ Confirm", style=discord.ButtonStyle.green)
```
StephenDaDev: same here, as far as I can think of, there is no use of symbols

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                    f"Do you want to create the snippet `{name}` with this attachment?",
+                    color=self.bot.main_color,
+                )
+                confirm_msg = await ctx.send(embed=embed, view=view)
+                await view.wait()
+
+                if confirmed is None or not confirmed:
+                    if confirmed is None:
+                        await confirm_msg.edit(
+                            content="⏱️ Timed out. Snippet not created.", view=None, embed=None
```
StephenDaDev: same

cogs/modmail.py
```diff
@@ -347,14 +393,128 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                    description="Failed to upload attachment. Please try again.",
+                )
+                return await ctx.send(embed=embed)
+
+        # Require at least text or attachment
+        if not value and not file_id:
+            embed = discord.Embed(
+                title="Error",
+                color=self.bot.error_color,
+                description="You must provide either text content or an attachment for the snippet.",
```
StephenDaDev: Usually our current convention for misformed is to send the help menu again

Image

cogs/modmail.py
StephenDaDev: agree

cogs/modmail.py
```diff
@@ -457,28 +620,114 @@ async def snippet_remove(self, ctx, *, name: str.lower):
...
+                attachment = ctx.message.attachments[0]
+
+                # Validate file size
+                max_size_mb = self.bot.config.get("snippet_attachment_max_size")
+                max_size_bytes = max_size_mb * 1024 * 1024
+                if attachment.size > max_size_bytes:
+                    embed = discord.Embed(
+                        title="Error",
+                        color=self.bot.error_color,
+                        description=f"Attachment exceeds the maximum file size of {max_size_mb} MB. "
```
StephenDaDev: It seems like if adding to snippet via an edit the large file warning is not displayed (logic in R449-R453).

cogs/modmail.py
```diff
@@ -457,28 +620,114 @@ async def snippet_remove(self, ctx, *, name: str.lower):
...
+
+            if attachment_info and old_file_id:
+                if not await self.bot.api.delete_snippet_attachment(old_file_id):
+                    logger.warning("Failed to delete old attachment for %s.", name)
+
+            description = f"`{name}` has been updated."
+            if value:
+                description += f'\nText: "{truncate(value, 100)}"'
+            if attachment_info:
+                description += f"\n📎 New attachment: `{attachment_info}`"
```
StephenDaDev: same comments on emoji use

core/config_help.json
```diff
@@ -856,6 +856,18 @@
       "See also: `anonymous_snippets`."
     ]
   },
+  "snippet_attachment_max_size": {
+    "default": "10 (MB)",
```
StephenDaDev: same comments on MB vs MiB

core/config_help.json
```diff
@@ -856,6 +856,18 @@
...
+    "description": "Maximum file size in megabytes (MB) for attachments when creating or editing snippets.",
+    "examples": [
+      "`{prefix}config set snippet_attachment_max_size 5` (5 MB)",
+      "`{prefix}config set snippet_attachment_max_size 20` (20 MB)"
+    ],
+    "notes": [
+      "Attachments larger than this size will be rejected when adding or editing snippets.",
+      "Value is specified in megabytes (MB)."
```
StephenDaDev: I think a note explaining the cons of raising this filesize without considering long term implications should be added

StephenDaDev (changes requested):
please see my mostly formatting consistency comments

cogs/modmail.py
```diff
@@ -116,6 +119,144 @@ def _resolve_user(self, user_str):
...
+                )
+            confirmed = True
+            await interaction.response.edit_message(view=None)
+            view.stop()
+
+        async def cancel_callback(interaction: discord.Interaction):
+            nonlocal confirmed
+            if interaction.user.id != ctx.author.id:
+                return await interaction.response.send_message(
+                    "Only the command author can cancel.", ephemeral=True
```
StephenDaDev: I think this should be an embed to be consistent with the rest of the bot, using error color

cogs/modmail.py
```diff
@@ -116,6 +119,144 @@ def _resolve_user(self, user_str):
...
+                )
+            confirmed = False
+            await interaction.response.edit_message(
+                content=f"Cancelled. Snippet not {past_tense}.", view=None, embed=None
```
StephenDaDev: same here

cogs/modmail.py
```diff
@@ -347,15 +541,85 @@ async def snippet_add(self, ctx, name: str.lower, *, value: commands.clean_conte
...
+                return await self._send_snippet_result(ctx, embed, confirm_msg)
+
+        if file_id and (self.bot.get_command(name) or name in self.bot.snippets or name in self.bot.aliases):
+            cleanup_failed = not await self._delete_snippet_attachment(file_id, name)
+            description = (
+                f"The name `{name}` became unavailable while the attachment was being uploaded. "
+                "Please try again with another name."
+            )
+            if cleanup_failed:
+                description += " The uploaded file could not be cleaned up."
```
StephenDaDev: how should users handle this when this happens?

StephenDaDev (dismissed):

cogs/utility.py
lorenzo132: This was to fix the help command, it was introduced somewhere, and it was a bug i found was small enough to include here.

StephenDaDev (dismissed):
Tested working, no further comments from me

martinbndr (dismissed):

cogs/modmail.py
```diff
@@ -250,13 +404,25 @@ async def snippet(self, ctx, *, name: str.lower = None):
...
-                val = self.bot.snippets[snippet_name]
+                snippet_data = self.bot.snippets[snippet_name]
+                snippet_text = self._get_snippet_text(snippet_data)
+                has_attachment = self._has_snippet_attachment(snippet_data)
+
+                if snippet_text:
+                    description = snippet_text
+                elif has_attachment:
+                    description = "This is an attachment-only snippet."
+                else:
```
sebkuip: In what case would this else be reached? A snippet must have an attachment or text right? It shouldn't be possible to make a snippet with no content.

core/thread.py
```diff
@@ -2078,6 +2109,26 @@ def lottie_to_png(data):
...
+                logger.warning(
+                    "Failed to convert snippet image %s to file.",
+                    getattr(a, "filename", "unknown"),
+                    exc_info=True,
+                )
+                continue
+
+            if not embedded_image:
+                embed.set_image(url=f"attachment://{a.filename}")
+                embed.add_field(name="Image", value=a.filename)
```
sebkuip: This is inconsistent with how reply works, where the filename is a link (top is snippet, bottom is reply)

Image

sebkuip (changes requested):
There are some issues still in the code. Please resolve these

PR #3425, opened 2025-12-21
fix: breaking threadmenu fixes.

lorenzo132:
This PR solves the following:
- thread creation menu null pointer bugs
- false cancelled cache entry
- If a menu times out, and the user send multiiple messages in the meantime the user would not be able to create a new thread with previous code.

### DIFF
--- a/core/clients.py
+++ b/core/clients.py
@@ -661,32 +661,52 @@ class MongoDBClient(ApiClient):
         channel_id: str = "",
         type_: str = "thread_message",
     ) -> dict:
-        channel_id = str(channel_id) or str(message.channel.id)
-        message_id = str(message_id) or str(message.id)
+        channel_id = str(channel_id) or (str(message.channel.id) if message else "")
+        message_id = str(message_id) or (str(message.id) if message else "")
 
-        data = {
-            "timestamp": str(message.created_at),
-            "message_id": message_id,
-            "author": {
-                "id": str(message.author.id),
-                "name": message.author.name,
-                "discriminator": message.author.discriminator,
-                "avatar_url": message.author.display_avatar.url if message.author.display_avatar else None,
-                "mod": not isinstance(message.channel, DMChannel),
-            },
-            "content": message.content,
-            "type": type_,
-            "attachments": [
-                {
-                    "id": a.id,
-                    "filename": a.filename,
-                    "is_image": a.width is not None,
-                    "size": a.size,
-                    "url": a.url,
-                }
-                for a in message.attachments
-            ],
-        }
+        if message:
+            data = {
+                "timestamp": str(message.created_at),
+                "message_id": message_id,
+                "author": {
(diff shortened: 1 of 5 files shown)


StephenDaDev:
I'll lump this in with v4.3.0 as well even though it wasn't included in our previous discussions

martinbndr (dismissed):

martinbndr (approved):

core/thread.py
```diff
@@ -2986,7 +3020,34 @@ async def callback(self, interaction: discord.Interaction):
...
+                                    )
+                                    # Ensure minimal attributes for Context if still missing (DummyMessage handles some, but we need more for commands)
+                                    if not synthetic._message:
+                                        # Identify a valid channel
+                                        ch = self.outer_thread.channel
+                                        if not ch:
+                                            # If channel isn't ready, we can't really invoke a command in it.
+                                            continue
+
+                                        from unittest.mock import MagicMock
```
sebkuip: Please don't make imports halfway into the file. Also is a unittest class really the only way to do this? This isn't the intended way to use this class.

core/thread.py
```diff
@@ -1059,7 +1060,10 @@ async def close(
         """Close a thread now or after a set time in seconds"""
 
         # restarts the after timer
-        await self.cancel_closure(auto_close)
+        await self.cancel_closure(
+            auto_close,
+            mark_auto_close_cancelled=not auto_close,
```
sebkuip, line 1061: I find this default value of `not auto_close` quite weird looking. Exactly what is this used for, and if it's based on the value of `auto_close`, can't you just use that everywhere?

core/clients.py
```diff
@@ -662,39 +662,52 @@ async def append_log(
...
+                "author": {
+                    "id": str(message.author.id),
+                    "name": message.author.name,
+                    "discriminator": message.author.discriminator,
+                    "avatar_url": (
+                        message.author.display_avatar.url if message.author.display_avatar else None
+                    ),
+                    "mod": not isinstance(message.channel, DMChannel),
+                },
+                "content": message.content,
```
sebkuip: In the old code, the bot would extract forwarded content. Now it's falling back to regular message.content, which would break if the message is a forward.

sebkuip (changes requested):

StephenDaDev:
Added to changelog for 4.3.0

sebkuip (dismissed):
After the changes requested have been resolved, this looks fine now.

StephenDaDev (dismissed):
Looks good and seems to function properly without error or sideeffects on test instance.

martinbndr (dismissed):

sebkuip (approved):
Still looking good.

StephenDaDev (approved):
me too

martinbndr (approved):

PR #3429, opened 2025-12-30
Fix threadmenu command execution

martinbndr:
This commit fixes the command execution for the threadmenu that caused commands to be sent to the recipient DM instead of the thread channel.
The root cause was that the source message (users DM) was used to generate the context for the commands being executed. Now it uses the thread genesis_message to create all needed info for the commands to be run. Responses will be forwarded to the thread channel.
Credits to the advanced menu plugin from @sebkuip which has realized it that way.

### DIFF
--- a/core/thread.py
+++ b/core/thread.py
@@ -2964,7 +2964,7 @@ class ThreadManager:
                                 # Create a synthetic message object that makes the bot appear
                                 # as the author for menu-invoked command replies so the user
                                 # selecting the option is not shown as a "mod" sender.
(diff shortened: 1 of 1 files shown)


lorenzo132 (approved):
Can confirm this resolves the issue after some testing.

martinbndr:
We should soon implement these changes with the other threadmenu fixes to make it easier for users with the transition.

sebkuip (approved):
Looks all good to me

PR #3430, opened 2026-01-02
Plugin registry checks upon start

martinbndr:
Plugins will be checked upon start if they are not in the registry and REGISTRY_PLUGINS_ONLY is set to Yes they will be removed. Also plugins are allowed to be removed regardless of being in the registry or not.

### DIFF
--- a/cogs/plugins.py
+++ b/cogs/plugins.py
@@ -153,6 +153,14 @@ class Plugins(commands.Cog):
                 logger.info("Migrated legacy plugin name: %s, now %s.", plugin_name, str(plugin))
                 self.bot.config["plugins"].append(str(plugin))
 
+            if self.bot.config.get("registry_plugins_only") and not plugin_name in self.registry:
+                self.bot.config["plugins"].remove(plugin_name)
+                logger.info(
+                    "Loading of plugin %s was skipped and the plugin was removed because it is not present in the registry.",
+                    plugin_name,
+                )
+                continue
+
             try:
                 await self.download_plugin(plugin)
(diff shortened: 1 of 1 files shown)


martinbndr:
Useful for:
- Plugins being removed from the registry (for example due to abuse)
- Hosters that run Modmail for other users and let them provide their DB. They could manually add the plugin to the db and the bot will install regardless of the REGISTRY_PLUGINS_ONLY .env var.

lorenzo132 (dismissed):

cogs/plugins.py
```diff
@@ -318,14 +326,15 @@ async def parse_user_input(self, ctx, plugin_name, check_version=False):
...
-            if self.bot.config.get("registry_plugins_only"):
-                embed = discord.Embed(
-                    description="This plugin is not in the registry. To install this plugin, "
-                    "you must set `REGISTRY_PLUGINS_ONLY=no` or remove this key in your .env file.",
-                    color=self.bot.error_color,
-                )
-                await ctx.send(embed=embed)
-                return
+            if check_registry:
+                if self.bot.config.get("registry_plugins_only"):
```
sebkuip: These if statemetns could be combined with an `and` for readability purposes.

sebkuip (changes requested):
A small change request for readability

sebkuip (approved):
LGTM

StephenDaDev (approved):
Looks good.

lorenzo132 (approved):
lgtm.

PR #3451, opened 2026-07-20
Add `thread_creation_menu_timeout` to `duration_seconds` config group

### LINKED ISSUE #3449: [BUG]: the config `thread_creation_menu_timeout` is not part of the `duration_seconds` group and thus is not validated to be valid
### Bot Version

V4.2.1

Other

### Error Logs

https://hastebin.cc/kinocewahu.sql

### Screenshots

_No response_

### Additional Information

See this line https://github.com/modmail-dev/Modmail/blob/49330b4a0eb1004af3085c0683e8d09608109277/core/config.py#L243

The config should be in this group such that it will be attempted to be converted and throws an error upon setting with wrong values.
This causes an error upon trying to use most threadmenu commands as it fetches an invalid config value

sebkuip:
This config value is not validated to be in seconds, but the threadmenu expects seconds. This PR adds this value to the `duration_seconds` config group such that it is validated, and if possible converted from other formats.
afbeelding

This closes #3449

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -240,7 +240,7 @@ class ConfigManager:
         "log_expiration",
     }
 
-    duration_seconds = {"snooze_default_duration"}
+    duration_seconds = {"snooze_default_duration", "thread_creation_menu_timeout"}
 
     booleans = {
         "use_user_id_channel_name",


StephenDaDev (approved):
tested working, resolves bug

lorenzo132 (approved):

martinbndr (approved):

PR #3453, opened 2026-08-03
Implement Slash Commands to Reduce Message Content Intent Reliance

lorenzo132:


martinbndr (approved):

StephenDaDev:
@lorenzo132 began to look over this and it looks like it still has the old version of your arg responses in it. I don't want to press update branch as I'm not sure what issues it'll cause. Can you please take a look at this? ty

